PR #8, opened 2026-04-12
feat: hardening — safety checks, kill switches, and self-deprecation

thepiper18:
## Summary

This PR adds 6 safety and operability features to make the interceptor robust against Claude Code internal changes and self-documenting about when it's no longer needed.

**Motivation:** The interceptor does important work, but it's tightly coupled to CC internals (fingerprint salt, block format, ordering conventions). If CC changes those internals, most fixes fail gracefully to no-ops — except the fingerprint rewrite, which can actively make cache performance *worse*. This PR eliminates that risk and adds controls for users to graduate from individual fixes as Anthropic ships upstream fixes.

### 1. Fingerprint round-trip safety check (P0)
Before rewriting `cc_version`, verify our salt/indices reproduce the fingerprint CC sent. If verification fails (CC changed its algorithm), skip the rewrite instead of corrupting it. Turns the only malignant failure mode into a safe no-op.

### 2. Master kill switch + per-fix toggles
- `CACHE_FIX_DISABLED=1` — bypass all bug fixes while keeping monitoring + optimizations active
- `CACHE_FIX_SKIP_{RELOCATE,FINGERPRINT,TOOL_SORT,TTL,IDENTITY}` — disable individual fixes

### 3. Persistent effectiveness stats
Track per-fix applied/skipped counts across sessions in `~/.claude/cache-fix-stats.json`. Relocate fix uses bug-presence detection (scanned N resume sessions, found 0 scattered blocks) rather than fix-application count for more reliable dormancy signals. Atomic writes (temp+rename), 30-day auto-prune.

### 4. Startup health status line
On first API call, logs per-fix status:
```
cache-fix health: relocate=active(2h ago) fingerprint=dormant(5 clean sessions) tool_sort=active ttl=active identity=waiting
```
Includes advisory messages for dormant fixes ("consider `CACHE_FIX_SKIP_RELOCATE=1`") and safety-blocked fixes.

### 5. Cache regression detector
In-memory ring buffer tracking `cache_read` ratio from API responses. If ratio drops below 50% across 5+ consecutive calls (turn 2+), warns the user — especially useful when fixes are disabled and CC regresses.

### 6. README documentation
New sections: "Graduating from Fixes" (three-purpose lifecycle table, health status guide, regression detection), "Safety" (fingerprint verification, fail-safe design guarantee).

## Design philosophy

The interceptor serves three purposes with different lifecycles:

| Purpose | Examples | Lifecycle |
|---------|----------|-----------|
| **Bug fixes** | Relocate, fingerprint, tool sort, TTL | Temporary bridge — obsolete when CC fixes upstream |
| **Monitoring** | Quota tracking, microcompact detection, GrowthBook flags | Permanent — detects future regressions |
| **Optimizations** | Image stripping, output efficiency rewrite | Permanent, opt-in |

`CACHE_FIX_DISABLED=1` disables the first category only. The kill switches and dormancy detection apply to bug fixes, not monitoring or optimizations.

The full safety lifecycle: **fixes active → dormancy detected → user disables fixes → monitoring-only → regression detector fires → re-enable suggestion**.

## Test plan

- [x] 75 tests pass (28 new + 47 existing), 0 failures
- [x] Fingerprint safety: stale salt detection, round-trip verification, edge cases (empty/short messages)
- [x] Kill switches: master switch, per-fix toggles, precedence rules
- [x] Stats: schema creation, counter increments, dormancy thresholds, bug-presence tracking
- [x] Health line: time formatting, status classification, full line generation
- [x] Regression detector: ratio computation, threshold detection, insufficient-history guard
- [x] All existing stabilizeFingerprint tests updated for round-trip verification
- [x] Manual testing with live Claude Code v2.1.104 session — verified fixes-on vs fixes-off:
  - Fixes ON: 99.5% avg cache hit rate (turn 2+), 0 full cache busts
  - Fixes OFF: 74.3% avg cache hit rate, 1 full cache bust (0% on resume)
  - Fingerprint safety check confirmed working: CC v2.1.104 changed salt (sent '3b6', we computed '38c'), rewrite correctly skipped
  - Block relocation fix active and needed on every resumed turn
  - Health line, stats file, and quota tracking all producing correct output

## Breaking changes

None. All new features are additive. Default behavior is unchanged — all fixes remain active unless explicitly disabled via env vars.

---

Built with analysis from a [3-cycle meta:think synthesis](https://github.com/thepiper18/claude-code-cache-fix/blob/feat/hardening/research/think-2026-04-12-claude-code-cache-fix-hardening.md) (89% convergence) covering Socratic questioning, first principles, second-order effects, pre-mortem, inversion, and Pareto prioritization. Research included Context7 docs on Node.js fetch interception patterns (Undici `.compose()`, MSW lifecycle).

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,67 @@ The module intercepts `globalThis.fetch` before Claude Code makes API calls to `
 
 All fixes are idempotent — if nothing needs fixing, the request passes through unmodified. The interceptor is read-only with respect to your conversation; it only normalizes the request structure before it hits the API.
 
+## Graduating from Fixes
+
+The interceptor serves three purposes with different lifecycles:
+
+| Purpose | Examples | When to disable |
+|---------|----------|-----------------|
+| **Bug fixes** | Block relocation, fingerprint, tool sort, TTL | When CC fixes the underlying bug — check the health line |
+| **Monitoring** | Quota tracking, microcompact detection, GrowthBook flags | Keep permanently — these detect future regressions |
+| **Optimizations** | Image stripping, output efficiency rewrite | Keep as long as they help your workflow |
+
+### Health status
+
+On first API call, the interceptor logs a health status line (requires `CACHE_FIX_DEBUG=1`):
+
+```
+cache-fix health: relocate=active(2h ago) fingerprint=dormant(5 clean sessions) tool_sort=active ttl=active identity=waiting
+```
+
+Status meanings:
+- **active(Xh ago)** — fix was applied recently
+- **dormant(N clean sessions)** — bug not detected in N resume sessions; CC may have fixed it
+- **safety-blocked(Nx)** — round-trip verification failed; CC changed its algorithm, fix auto-disabled
+- **waiting** — fix hasn't been triggered yet
+
+When a fix shows `dormant`, you can safely disable it:
+```bash
+export CACHE_FIX_SKIP_RELOCATE=1  # example
+```
+
+To disable all fixes but keep monitoring:
+```bash
+export CACHE_FIX_DISABLED=1
+```
+
+### Regression detection
+
+If cache_read ratio drops below 50% across 5+ calls after disabling fixes, you'll see:
+```
+REGRESSION WARNING: cache_read ratio averaged 12% across last 5 calls.
+Fixes are disabled — consider re-enabling to recover cache performance.
+```
+
+## Safety
+
+### Fingerprint round-trip verification
+
+Before rewriting the `cc_version` fingerprint, the interceptor verifies that its
+hardcoded salt and character indices reproduce the fingerprint Claude Code sent.
+If verification fails (CC changed its algorithm), the rewrite is skipped automatically.
+This ensures the interceptor can never make cache performance *worse* than stock CC.
+
+### Fail-safe design
+
+Every fix is designed to fail to a no-op:
+- If block detection regexes don't match → blocks aren't relocated (CC behavior)
+- If fingerprint format changes → fingerprint isn't rewritten (CC behavior)
+- If tool sort produces no changes → payload passes through untouched
+- If TTL injection target structure changes → TTL isn't injected (CC behavior)
+
+The interceptor can only *help* or *do nothing*. It cannot make things worse.
+
 ## Image stripping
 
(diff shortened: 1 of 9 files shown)


cnighswonger:
@thepiper18 — this is a substantial and well-designed contribution. The fingerprint round-trip safety check in particular addresses a real gap: if Anthropic changes their fingerprint algorithm, our current rewrite could actively make cache performance worse rather than failing to a no-op. That's been an unaddressed risk since v1.0.0 and your verification-before-rewrite approach is the right fix.

The "three purposes with different lifecycles" framework (bug fixes / monitoring / optimizations) and the `CACHE_FIX_DISABLED=1` master switch that only disables the first category — that's the right architecture for where this project is heading. Several of the fixes are already showing dormancy in some users' environments as Anthropic ships incremental improvements, so per-fix toggles and dormancy detection are timely.

Two notes for the review pass Monday:

1. **Fingerprint round-trip logic** — this is the most safety-critical change in the PR and we want to walk through the verification path carefully. If the round-trip check has a false-positive (incorrectly concludes CC changed the algorithm when it didn't), it would silently disable the fingerprint fix for users who still need it. We'll look at the edge cases.

2. **Portuguese guide** — great addition for i18n reach (we currently have EN + ZH). The guide references your fork (`thepiper18/claude-code-cache-fix/tree/feat/hardening`) for install instructions — that'll need to point at the main repo before merge.

Your independent benchmark (99.5% vs 74.3%, confirmed on v2.1.104) and the 28 new tests bringing the suite to 75 are both strong signals. Planning to review and merge Monday — if clean this ships as v1.8.0 given the scope.

cnighswonger:
Merged and shipped as v1.8.0 — live on npm now.

```bash
npm install -g claude-code-cache-fix@1.8.0
```

Your Portuguese guide is linked from the README header alongside the English and Chinese versions. The fork references in the guide have been updated to point at the main repo.

One small change during merge: README conflict resolution placed the "Graduating from Fixes" and "Safety" sections before the "Status line" section (both were added independently to the same location). The content is unchanged.

The timing on this was critical — v2.1.104 has partially converged on some of our bug fixes upstream, and your dormancy detection is now exactly what users need to see which fixes are still active vs. which CC has handled. We published [Part 2 of our blog series](https://veritassuperaitsolutions.com/the-three-layer-gate/) today referencing the lifecycle model from this PR.

Thanks for the contribution — the safety-first design philosophy is exactly right for a tool that patches someone else's internals.

PR #30, opened 2026-04-17
Add reminder_strip — remove bookkeeping system-reminder blocks entirely

deafsquad:
Complements `smoosh_normalize` / `smoosh_split` by **removing** purely-bookkeeping system-reminder blocks from user messages entirely, instead of stabilizing them in place.

## What it targets

Per-turn CC-internal reminders whose values the agent doesn't need to condition behavior on:

- `Token usage: <N>/<M>; <K> remaining`
- `Output tokens — turn: <X> · session: <Y>`
- `USD budget: $<X>/$<Y>; $<Z> remaining`
- `The task tools haven't been used recently. …`
- `The TodoWrite tool hasn't been used recently. …`
- `Remaining conversation turns: <N>`
- `Messages until auto-compact: <N>`

Each pattern is anchored with `^` so bookkeeping phrases embedded inside a larger semantic reminder are NOT stripped (false-positive guard — e.g., if a user discusses the concept mid-sentence or an agent report echoes the phrase).

## How it composes with the existing passes

- `smoosh_normalize` stabilizes dynamic values in-place for 4 known patterns → agent still sees them
- `smoosh_split` peels smooshed reminders back into standalone text blocks
- `reminder_strip` (this) runs AFTER `smoosh_split` and removes the matched blocks outright

The order matters: split makes the blocks visible as standalone text; strip then drops the bookkeeping subset.

## What it deliberately does NOT touch

Hook-injected reminders pass through untouched:

- `PreToolUse:X hook blocking error …`
- `PostToolUse:X hook additional context …`
- `UserPromptSubmit hook additional context …`
- `SessionStart:X hook success …` (handled by separate `session_start_normalize` extension PR)
- `The user sent a new message while you were working …`
- `thinking-enrichment`, `action-tracker`, `classify-user-intent`, custom user hooks

The agent genuinely needs these visible on the turn they fire. I tried a history-only-strip approach earlier and abandoned it — shifting "last user message" semantics create new drift at the message boundary every turn. Leaving hook reminders untouched is the safer choice, and their residual drift is small compared to bookkeeping churn.

## Conventions matched

- Inline filter in `preload.mjs`, no side effects outside the content array
- `isBookkeepingReminder(text): boolean` — pure predicate, exported for unit tests
- `shouldApplyFix("reminder_strip")` gate
- `recordFixResult("reminder_strip", "applied"|"skipped")`
- `_STATS_SCHEMA` entry: `{ applied: 0, skipped: 0, lastApplied: null }`
- Default ON; opt-out via `CACHE_FIX_SKIP_REMINDER_STRIP=1`

## Tests

14 new cases in `test/isBookkeepingReminder.test.mjs`:

- All 7 enumerated patterns match (including em-dash + middot format for output tokens, both `Message` and `Messages` for auto-compact counter)
- Hook-injected reminders (5 variants) are NOT matched
- claudeMd preamble is NOT matched
- Deferred-tools attachment is NOT matched
- Plain user text / missing wrapper is NOT matched
- Non-string inputs return false without throwing
- Partial-body pattern match is rejected (anchor check)
- Off-format values (missing semicolon, non-numeric counts) are rejected

Full suite: **89 passing, 0 failing** (14 new + 75 existing).

## Empirical validation

Part of the stack that took first-request `cache_creation_input_tokens` on `--continue` from **940,251** → **1,704** in a long-running resume-heavy session. On that session it fires 20–60+ times per request as bookkeeping blocks accumulate in history.

Related: cnighswonger/claude-code-cache-fix#49585 (discussion thread).

### DIFF
--- a/preload.mjs
+++ b/preload.mjs
@@ -388,6 +388,65 @@ function normalizeSessionStartText(text) {
   return [out, count];
 }
 
+// --------------------------------------------------------------------------
+// Bookkeeping-reminder strip
+// --------------------------------------------------------------------------
+//
+// Complements `smoosh_normalize` / `smoosh_split`: where normalize stabilizes
+// bytes in-place and split peels smooshed reminders back into standalone
+// text blocks, this pass REMOVES purely-bookkeeping reminder blocks entirely
+// from the outbound body. Zero model visibility, zero drift.
+//
+// Targeted patterns (all CC-internal, per-turn values the agent doesn't need
+// to condition behavior on):
+//   - `Token usage: <N>/<M>; <K> remaining`
+//   - `Output tokens — turn: <X> · session: <Y>`
+//   - `USD budget: $<X>/$<Y>; $<Z> remaining`
+//   - `The task tools haven't been used recently. …`
+//   - `The TodoWrite tool hasn't been used recently. …`
+//   - `Remaining conversation turns: <N>`
+//   - `Messages until auto-compact: <N>`
+//
+// Hook-injected reminders (thinking-enrichment, action-tracker,
+// PreToolUse/PostToolUse blocking errors, UserPromptSubmit additional
+// context, custom user hooks) are deliberately NOT stripped here — the
+// agent needs that feedback visible in the turn it fires, and attempting a
+// history-only filter creates per-turn drift of its own (the "last user
+// message" shifts each turn, so a reminder preserved at turn N gets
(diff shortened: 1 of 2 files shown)


cnighswonger:

cnighswonger:
Content landed in `6e2448d`, released in v2.0.0. Closing for housekeeping — the extension is live and running in production.

PR #31, opened 2026-04-17
Add cache_control_normalize — pin marker at canonical position to stop per-turn drift

deafsquad:
Pin the `cache_control` marker at a single canonical position so CC's natural marker-shuffle between turns stops invalidating the previous-last-block's cached bytes.

## Problem

Anthropic's prompt-cache uses `cache_control: {type: "ephemeral", ttl: ...}` markers on content blocks as cache breakpoints. CC places this marker on "the last block of the last user message" each turn — which shifts as new turns arrive.

Turn N:
```json
{"type": "tool_result", "content": "...", "tool_use_id": "a",
 "cache_control": {"type": "ephemeral", "ttl": "1h"}}
```

Turn N+1 (same message, now earlier in history):
```json
{"type": "tool_result", "content": "...", "tool_use_id": "a"}
```

The `cache_control` field silently disappears from what WAS the last-block last turn. 53 bytes of drift in that block's JSON → server can't partial-hit against its cached version of the prefix ending at that block. On a 200+ message session this is a few thousand to tens of thousands of cache-creation tokens per turn, depending on what other drift compounds at the same position.

## Fix

Enforce a canonical position on every outbound body:

1. Strip every existing `cache_control` marker from user-message content blocks
2. Place a single `{type: "ephemeral", ttl: "1h"}` marker on the LAST block of the LAST user message

Idempotent. Fast path: if the canonical block already has the correct marker AND it's the only user-side marker, the body is left byte-untouched.

## What's NOT touched

System-side markers (e.g., on `system[2]` for the global prompt) pass through — they're CC's stable breakpoint for the system prompt and work correctly as-is.

## Conventions matched

- Inline in `preload.mjs`, no side effects outside the call site
- `stripCacheControlMarkers(msg)` → number — pure helper, exported
- `countUserCacheControlMarkers(body)` → number — pure helper, exported
- `CACHE_CONTROL_CANONICAL_MARKER` — exported constant
- `shouldApplyFix("cache_control_normalize")` gate
- `recordFixResult("cache_control_normalize", "applied"|"skipped")`
- `_STATS_SCHEMA` entry: `{ applied: 0, skipped: 0, lastApplied: null }`
- Default ON; opt-out via `CACHE_FIX_SKIP_CACHE_CONTROL_NORMALIZE=1`
- **Runs LAST** in the extension pipeline so the canonical position is calculated against the final content array (after `smoosh_split`, `reminder_strip` etc. potentially mutate block counts)

## Tests

12 new cases in `test/cacheControlNormalize.test.mjs`:

**`stripCacheControlMarkers`:**
- Single marker stripped, text/type preserved
- No markers → returns 0
- Multiple markers in same message stripped
- Assistant role: skipped (marker preserved)
- String content: skipped
- Null/empty/missing inputs: 0 without throwing
- Non-`cache_control` block fields preserved

**`countUserCacheControlMarkers`:**
- Counts across user messages only (assistant markers ignored)
- Zero when none present
- Null/empty safety
- String-content user message doesn't crash

**`CACHE_CONTROL_CANONICAL_MARKER`:**
- Constant shape: `{type: "ephemeral", ttl: "1h"}`

Full suite: **87 passing, 0 failing** (12 new + 75 existing).

## Empirical validation

In our long-running resume-heavy session, the fast path hits ~100% of the time — CC places its own marker at the same canonical position, so this pass is a no-op on every request. It's a **safety net**: when CC diverges (first post-resume request, post-compaction, or edge cases in specific tool-use shapes), the pin kicks in and the previous-last-block's bytes stay stable.

Part of the stack contributing to the 940K → 1.7K first-request `cache_creation_input_tokens` reduction on `--continue`.

## A note on my earlier A/B

During development I saw a pid with consistent ~200K cw-per-request where this extension was active. Offline repro on captured bodies proved it was not the cause — the pipeline was fully idempotent and reported zero mutations. The misses correlated with rapid-fire request bursts within 13 seconds of each other and resolved autonomously after a short idle period, pointing to transient Anthropic-side cache-write propagation latency rather than an extension bug. Flagging here in case you've seen similar during your own validation.

Related: cnighswonger/claude-code-cache-fix#49585 (discussion thread).

### DIFF
--- a/preload.mjs
+++ b/preload.mjs
@@ -388,6 +388,70 @@ function normalizeSessionStartText(text) {
   return [out, count];
 }
 
+// --------------------------------------------------------------------------
+// cache_control marker position-normalizer
+// --------------------------------------------------------------------------
+//
+// Anthropic's prompt-cache uses `cache_control: {type: "ephemeral", ttl: ...}`
+// markers on content blocks as cache breakpoints. CC places this marker on
+// "the last block of the last user message" each turn — which shifts as new
+// turns arrive. When the marker moves, the PREVIOUS last-block's JSON loses
+// the cache_control field → that block's bytes differ from the server's
+// cached version → partial re-cache on top of the stable system-prompt
+// cache.
+//
+// Enforce a canonical position on every outbound body:
+//   1. Strip every existing cache_control marker from user-message content
+//      blocks.
+//   2. Place a single {type: "ephemeral", ttl: "1h"} marker on the LAST
+//      content block of the LAST user message.
+//
+// Fast path: if the canonical block already has the correct marker AND it's
+// the only user-side marker, the body is left untouched — ensures the pass
+// is a true no-op when nothing changed.
+//
+// System-side markers (e.g., on `system[2]` for the global prompt) are NOT
+// touched — they're CC's stable breakpoint for the system prompt and work
+// correctly.
+// --------------------------------------------------------------------------
+
+const CACHE_CONTROL_CANONICAL_MARKER = { type: "ephemeral", ttl: "1h" };
+
+/**
+ * Strip every cache_control marker from a single user message's content
+ * blocks. Returns the number stripped. Mutates the message's content array
+ * in place.
+ */
+function stripCacheControlMarkers(msg) {
+  if (!msg || msg.role !== "user" || !Array.isArray(msg.content)) return 0;
+  let n = 0;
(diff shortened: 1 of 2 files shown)


cnighswonger:

cnighswonger:
Content landed in `dfeac92`, released in v2.0.0. Closing for housekeeping — the extension is live and running in production.

PR #32, opened 2026-04-17
Add tool_use_input_normalize — prevent tool_use.input field-set drift between turns

deafsquad:
## Summary

Adds a new `tool_use_input_normalize` extension that walks every assistant-role message's `tool_use` blocks, looks up the tool's declared `input_schema.properties` from `body.tools`, and rewrites `input` to contain ONLY the schema keys in schema declaration order. This pins the serialization of `tool_use.input` to the schema's field set and key order so CC's own drift between turns can't break cache.

## The bug

CC's serialization of `tool_use.input` can drift between turns when the caller passed fields not declared in the tool's `input_schema.properties`. The pre-miss body may serialize only the schema keys (e.g. `{to, summary, message}`) while the post-miss body serializes the full caller-supplied set (e.g. `{to, summary, message, type, recipient, content}` — same tool_use_id, same turn position). That byte drift at a mid-history assistant message re-caches every block from that message forward → full-session-cost miss.

### Concrete observed instance

A SendMessage `tool_use` block on `msg[844]` drifted by **2334 bytes** between turns, causing a **619,722 `cache_creation_input_tokens`** miss at **15:16:52 UTC** on a long-running session:

- Tool's declared schema: 3 properties (`to`, `summary`, `message`)
- Caller supplied: 6 keys (`to`, `summary`, `message`, `type`, `recipient`, `content`)
- Pre-miss serialization: 3 schema-only keys
- Post-miss serialization: 6 keys (extras preserved)
- Result: cache break from `msg[844]` forward

## The fix

- Inline helper `normalizeToolUseInputsInBody(body) → number` in `preload.mjs`
- Builds `toolSchemas = { name: Set<keys> }` from `body.tools` entries with `input_schema.properties`
- Walks `body.messages`, only `role === "assistant"` with array content
- For each `type === "tool_use"` block, if the tool's schema is known and `input` has extra keys, rewrites `input` using only schema keys in **schema declaration order** (not input's original order)
- Tools whose schema is absent from `body.tools` are left untouched — we can't tell what's legitimate vs extra
- Pure in-place mutation returning a count; safe to call repeatedly (idempotent)

Runs AFTER `session_start_normalize` in the pipeline so drifted bytes are pinned before any downstream pass (`smoosh_*`, fingerprint, ttl) hashes the same block.

**Default ON**, opt-out via `CACHE_FIX_SKIP_TOOL_USE_INPUT_NORMALIZE=1`.

Agent behavior is unaffected — the extras weren't in the tool's declared schema so downstream consumers have no contract on them.

## Test plan

- [x] No-schema no-op (tool absent from `body.tools`)
- [x] No-extras no-op (input keys ⊆ schema keys)
- [x] Extras stripped (input keys ⊋ schema keys)
- [x] Multiple `tool_use` blocks in one message
- [x] User-role guard (user message with `tool_use`-shaped content untouched)
- [x] Missing `body.tools` guard (no-op)
- [x] Schema-order key preservation (NOT input's original order)
- [x] Non-array content robustness (role=user with string content doesn't crash)
- [x] Malformed body guards (missing messages / missing tools)
- [x] Idempotency (second call is a no-op)
- [x] Cross-turn byte stability regression
- [x] Defensive handling of missing/array/null `input`
- [x] Real captured-body regression fixture matching the observed SendMessage miss (6 caller keys → 3 schema keys)

All **98 suite tests pass** (85 prior + 13 new).

### DIFF
--- a/preload.mjs
+++ b/preload.mjs
@@ -388,6 +388,106 @@ function normalizeSessionStartText(text) {
   return [out, count];
 }
 
+// --------------------------------------------------------------------------
+// tool_use.input field-set normalization
+// --------------------------------------------------------------------------
+//
+// CC's serialization of `tool_use.input` can drift between turns when the
+// caller passes fields not declared in the tool's `input_schema.properties`.
+// Observed case: a SendMessage tool call where the caller passed
+// `{to, summary, message, type, recipient, content}`. Pre-miss body
+// serialized input as `{to, summary, message}` (3 schema-only keys).
+// Post-miss body (same tool_use_id, same turn position) serialized the
+// same block as `{to, summary, message, type, recipient, content}` (6 keys
+// — extras preserved). That byte drift at a mid-history assistant message
+// re-caches every block from that message forward → full-session-cost miss.
+//
+// Concrete instance: 2334-byte drift on ONE assistant-side tool_use block
+// caused a 619,722 `cache_creation_input_tokens` miss at 15:16:52 UTC on
+// msg[844] of a long-running session.
+//
+// This helper walks every assistant-role message's tool_use blocks, looks
+// up the tool's declared `input_schema.properties` from `body.tools`, and
+// rewrites `input` to contain ONLY the schema keys (in schema declaration
+// order). Tools with no schema in `body.tools` are left untouched — we
+// can't determine what's legitimate vs extra.
+//
+// Agent behavior is unaffected — extras weren't declared in the schema so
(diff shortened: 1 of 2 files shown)


cnighswonger:

cnighswonger:
Content landed in `ed4aac9`, released in v2.0.0. Closing for housekeeping — the extension is live and running in production.

PR #33, opened 2026-04-17
Add cache_control_sticky extension — preserve historical marker positions

deafsquad:
## What

Adds a new `cache_control_sticky` extension that preserves historical `cache_control` marker positions across turns by reinstating them on future outbound bodies from a per-session state file.

### The miss this fixes

Observed empirically on 2026-04-17 at 16:27:13 UTC: a 1284-message session emitted `cache_creation_input_tokens=804,428` (hit rate 2.3%). Diff of main-session bodies 585 vs 587 showed ONE message diverged — `msg[1281]` — which lost its `cache_control: {type: "ephemeral", ttl: "1h"}` marker (~43 bytes of JSON framing). CC had moved the only message-level marker onto the new last user message `msg[1283]`, dropping it from `msg[1281]`. That tail-of-message byte drift invalidates every downstream cached block.

### Why `cache_control_normalize` can't fix this

`cache_control_normalize` (the sibling extension) strips all user-side markers and pins exactly ONE canonical marker at the last block of the last user message. That solves the current-marker-drift class — but CC has already DROPPED the historical marker by the time the body reaches this preload. Normalize sees a single marker on the last msg and nothing to restore.

### Approach

Maintain per-session state tracking where markers have appeared. Reinstate them on future turns as additive preservation — up to 3 historical positions (Anthropic's hard total-marker cap is 4: `system[2]` + 3 message-level breakpoints).

Messages are identified by a **stable hash** so compaction rewrites and block-insertion passes (like `smoosh_split`) don't confuse the tracker:
- Prefer `sha1(role | tool_use_id)` when a `tool_use` or `tool_result` id is available.
- Otherwise hash `role | firstTextBlock.slice(0, 256)`.

On each outbound body:
1. Walk user messages; observe current cache_control markers and their message hashes.
2. Merge into prior state (keep unseen prior entries, refresh re-seen ones, LRU cap at 3).
3. For each tracked hash still present in this body, if the message carries no marker, emit a mutation to add one at its last block.
4. Write updated state atomically to `~/.claude/cache-fix-state/cache-control-sticky-<sha1(cwd)[:16]>.json`.

### Pipeline order

Runs AFTER `cache_control_normalize` (when both are enabled). Normalize places the canonical tail marker first; sticky re-adds historical ones on their hashed messages. Sticky skips any message already carrying a marker, so it's safe either way.

## Default on, opt-out

Default ON. Opt-out via `CACHE_FIX_SKIP_CACHE_CONTROL_STICKY=1` (matches the extension-family convention).

## Exports

Pure helpers exported for unit tests:
- `computeStickyMessageHash(msg)`
- `updateCacheControlStickyState(body, priorState)` — **pure**, returns `{newState, mutations}`
- `applyCacheControlSticky(body, key)` — wrapper with I/O
- `readCacheControlStickyState(key)`, `writeCacheControlStickyState(key, state)`
- `cacheControlStickyStatePath(key)`
- `CACHE_CONTROL_STICKY_MAX_POSITIONS` (= 3)
- `CACHE_CONTROL_STICKY_DEFAULT_MARKER` (= `{type: "ephemeral", ttl: "1h"}`)

## Tests

14 new cases covering:
- No marker-able messages → no-op, state empty
- First body with a marker → state records hash
- Subsequent body where CC removed the marker → marker re-added at last block
- LRU cap at 3 when 4 markers observed
- Hash stable across content-block insertions (smoosh_split compat)
- Hash stable across text changes when `tool_use_id` is the anchor
- Corrupt state file → empty state fallback, no throw
- Composes cleanly with canonical-last-msg marker
- Idempotent (second pass emits no new mutations)
- Different project keys → different state file paths
- End-to-end `applyCacheControlSticky` writes state and mutates body
- Body with no messages → returns 0, no crash
- `computeStickyMessageHash` returns null for uncontented messages
- Default marker constant shape

Full suite: **121 pass** (107 prior + 14 new).

## Files touched

- `preload.mjs` — +~290 lines: new section between `findDeferredToolsBlockInBody` and the core relocate fix; pipeline block after `deferred_tools_restore`, before `ttl`; `_STATS_SCHEMA` entry; export block additions.
- `test/cacheControlSticky.test.mjs` — new, 14 tests.

## Trade-offs

- State file is best-effort: corrupt JSON → fall back to empty, warn-log. Never throws.
- We only re-add when the message carries NO existing marker — won't overwrite normalize's canonical placement or user/tool-placed markers.
- Hash collision space is 2^64 (16 hex chars). Worst-case false-match → one message gets a spurious marker added; no cache correctness issue because markers are cumulative-safe up to the cap.

### DIFF
--- a/preload.mjs
+++ b/preload.mjs
@@ -504,6 +504,262 @@ function findDeferredToolsBlockInBody(body) {
   return null;
 }
 
+// --------------------------------------------------------------------------
+// cache_control_sticky — preserve historical marker positions across turns
+// --------------------------------------------------------------------------
+//
+// Covers a cache-miss class that cache_control_normalize can't reach by
+// itself. CC maintains at most one user-side cache_control marker at a time:
+// as conversation grows, CC moves the marker from the tail of one user turn
+// to the tail of the next, DROPPING it from the previous position. The
+// dropped position's block loses the ~43 bytes of `"cache_control":{"type":
+// "ephemeral","ttl":"1h"}` framing — a tail-of-message byte diff that
+// invalidates every downstream cached block (~600K tokens' worth on a
+// long-running session).
+//
+// Observed instance: at 16:27:13 UTC today, a 1284-message session emitted
+// cw=804,428 (hit=2.3%). Diff of main-session bodies 585 → 587 showed ONE
+// message diverged — msg[1281] — which lost its cache_control marker (43
+// bytes) because CC had moved the marker to the new last user msg[1283].
+//
+// cache_control_normalize places exactly ONE canonical marker at the last
+// block of the last user message on every outbound body. That solves the
+// current-marker-drift class but cannot preserve historical markers — CC
+// has already dropped them by the time the payload reaches this extension.
+//
+// This sticky extension maintains per-session state tracking where markers
+// have appeared in prior turns, and reinstates them on future turns as
+// additive preservation. Up to 3 historical message-level markers are
+// tracked (Anthropic's hard limit is 4 cache_control markers total — 1 for
+// system[2] + 3 for message-level breakpoints). When a historical position
+// would exceed the cap, the oldest tracked entry is dropped (LRU).
+//
+// Messages are identified by a stable hash so that compaction rewrites /
+// index shifts don't confuse the tracker:
+//   - If the message has a tool_use or tool_result block with an `id` or
+//     `tool_use_id`, hash `role|id`.
+//   - Otherwise hash `role|firstTextContent.slice(0, 256)`.
+//
+// Pipeline order: runs AFTER cache_control_normalize (when it's present) so
+// normalize first pins the canonical marker at the last user msg, then
+// sticky re-adds historical markers on their hashed messages. Skips any
+// message already carrying a marker (fast no-op when sticky fires first).
+//
+// Opt-out via CACHE_FIX_SKIP_CACHE_CONTROL_STICKY=1 (defaults ON).
+// --------------------------------------------------------------------------
+
+const CACHE_CONTROL_STICKY_DIR = join(homedir(), ".claude", "cache-fix-state");
+const CACHE_CONTROL_STICKY_MAX_POSITIONS = 3;
+const CACHE_CONTROL_STICKY_DEFAULT_MARKER = { type: "ephemeral", ttl: "1h" };
+
+/**
+ * Build the absolute state-file path for a given project key. Exported so
+ * tests can assert on path derivation without duplicating hash logic.
+ */
+function cacheControlStickyStatePath(key) {
+  const hash = createHash("sha1").update(String(key)).digest("hex").slice(0, 16);
+  return join(CACHE_CONTROL_STICKY_DIR, `cache-control-sticky-${hash}.json`);
+}
+
+/**
+ * Compute a stable hash identifier for a message that survives content-
+ * block insertions (e.g. smoosh_split peeling a reminder into a new block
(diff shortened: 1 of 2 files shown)


cnighswonger:
**Heads up — we hit a production issue with this PR and shipped a hotfix in v2.0.2.**

On resume, `cache_control_sticky` with `MAX_POSITIONS=3` + `cache_control_normalize` (1 canonical) + CC's system[2] marker = 5 `cache_control` blocks. Anthropic's hard limit is 4. Result: `400 invalid_request_error` on every API call once the session has enough history to fill all 3 sticky slots.

```
API Error: 400 {"type":"error","error":{"type":"invalid_request_error",
"message":"A maximum of 4 blocks with cache_control may be provided. Found 5."}}
```

**Fix (v2.0.2):** Reduced `MAX_POSITIONS` from 3 to 2. Budget: 1 system + 1 canonical (normalize) + 2 historical (sticky) = 4.

The interaction between #31 and #33 is the root cause — both are yours and both work correctly in isolation, but combined they exceed the marker cap. We should have caught this in review.

Updated test, comments, and changelog. Already on npm as `2.0.2`.

Going forward: per the new [CLAUDE.md](https://github.com/cnighswonger/claude-code-cache-fix/blob/main/CLAUDE.md), could you rebase future PRs against current main before requesting review? The cherry-pick conflict resolution we've been doing is error-prone on a single 2000+ line file. Clean fast-forward merges are safer for everyone.

cnighswonger:

cnighswonger:
Content landed in `6ae092d`, released in v2.0.1 (with a follow-up fix in v2.0.2 for a 4-marker-limit bug in the sticky logic). Closing for housekeeping — the extension is live and running in production.

deafsquad:
Own goal on my side — the #31 × #33 × CC-system-marker interaction was exactly the kind of cross-extension composition I should have stress-tested in the PR and didn't. Specifically: I tested sticky in isolation against a body without any existing `cache_control` markers, so the 3-slot budget looked fine. Running it against a real body where `cache_control_normalize` had already placed the canonical + CC had its system[2] = instant 5-block violation.

Thanks for catching it in production, shipping v2.0.2 and then v2.0.3 back-to-back, and calling out the dynamic-count design — that's the right long-term shape since Anthropic's per-config marker count can drift (CC 2.1.112 using 3 already proves the point). Hardcoded budget assumptions are the wrong axis to compute on.

**On the rebase workflow** — fully on board. I read your CLAUDE.md and it's exactly the discipline this shape of codebase needs (single 2000+ line file, cherry-pick risk is high, sequential merges keep conflicts local). For future PRs I'll:

1. Always rebase against current `upstream/main` before requesting review
2. Test on the exact commit I push (not "it passed a few commits ago")
3. `npm pack --dry-run` before any publish-adjacent work (I'm not publishing, but the principle applies if I ever need to verify what the tarball would contain against my branch)
4. Never ask you to resolve conflicts — if a PR conflicts, I'll handle the rebase

On your side: no shame in the review miss. The interaction surface between stateful extensions (sticky reads state, normalize writes state) is inherently tricky and our test coverage didn't include a "post-normalize + post-sticky = too many markers" assertion. I'll add that pattern to future PRs — a `test_pipeline_result_respects_anthropic_cache_control_cap` that counts markers across the mutated body after every extension has run.

On my side locally: swapped our vendored copy to v2.0.3 this afternoon. Verified the dynamic-count logic runs clean across our 1400+ message session. No more 5-marker bodies going out.

Thanks again for the rapid hotfix. Closing this PR's discussion here — content is already shipped via direct commits to main per your workflow.

Issue #52, opened 2026-04-22
Windows: pipeline.mjs passes raw drive-letter path to dynamic import(), all 8 cache-fix extensions fail to load → proxy is silently a passthrough


X-15:
## Bug

On Windows, **none of the proxy pipeline extensions load**, so the proxy degrades to a passthrough with zero cache fixes applied. Affects v3.0.x.

`proxy/pipeline.mjs` line 19:

```js
const mod = await import(join(dir, file) + "?t=" + Date.now());
```

`join(dir, file)` returns a raw Windows path like `C:\Users\...\proxy\extensions\fingerprint-strip.mjs`. Node's dynamic `import()` rejects that:

```
[pipeline] failed to load fingerprint-strip.mjs: Only URLs with a scheme in: file, data, node, and electron are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'c:'
[pipeline] failed to load cache-control-normalize.mjs: ... Received protocol 'c:'
[pipeline] failed to load cache-telemetry.mjs: ... Received protocol 'c:'
[pipeline] failed to load fresh-session-sort.mjs: ... Received protocol 'c:'
[pipeline] failed to load identity-normalization.mjs: ... Received protocol 'c:'
[pipeline] failed to load request-log.mjs: ... Received protocol 'c:'
[pipeline] failed to load sort-stabilization.mjs: ... Received protocol 'c:'
[pipeline] failed to load ttl-management.mjs: ... Received protocol 'c:'
proxy listening on 127.0.0.1:9801
```

The proxy starts and answers `/health` 200, so it looks healthy, but every request flows through unchanged. `cache-fix-stats.json` (preload mode) shows fixes happening; the proxy's pipeline never runs.

## Repro

- Windows 10, 11, or Server 2025
- Any Node 18+ (tested on Volta-managed 22.22.0 and stock npm)
- `npm install -g claude-code-cache-fix@latest`
- `node "$(npm root -g)/claude-code-cache-fix/proxy/server.mjs"` (or trigger via the VS Code extension)
- Stderr immediately prints the eight load failures shown above.

## Fix (one-line)

Wrap with `pathToFileURL`:

```diff
 import { readdir, readFile } from "node:fs/promises";
 import { join } from "node:path";
+import { pathToFileURL } from "node:url";
 ...
-      const mod = await import(join(dir, file) + "?t=" + Date.now());
[...]

PR #54, opened 2026-04-23
Honor HTTPS_PROXY / NO_PROXY / NODE_EXTRA_CA_CERTS / CACHE_FIX_PROXY_CA_FILE in proxy upstream forwarder

X-15:
## Summary

Make the proxy a well-behaved citizen behind corporate egress proxies and SSL-inspecting MITM CAs by wiring three industry-standard env vars (`HTTPS_PROXY`, `HTTP_PROXY`, `NO_PROXY`) plus one new package-specific knob (`CACHE_FIX_PROXY_CA_FILE`). With no env vars set, behavior is **unchanged** from prior versions (Node default agent, system trust store).

Tested end-to-end on Windows Server 2025 + plain npm and Windows 11 + Volta-managed Node 22.22.0, where v3.0.2 currently fails to forward (timeout / TLS-chain error) until these env vars are honored.

## Why

`proxy/upstream.mjs` (line 71 on main, before this change) calls `transport.request(options, ...)` with no `agent`. Consequences in a corporate environment:

- **`HTTPS_PROXY` / `HTTP_PROXY` / `NO_PROXY` are silently ignored.** Node's stdlib doesn't honor those without an explicit agent. Outbound `:443` connect to `api.anthropic.com` either times out or is blocked at the firewall, and the user gets no useful error.
- **Windows system proxy is ignored.** Node never reads the IE/winhttp proxy.
- **`NODE_EXTRA_CA_CERTS`** *is* honored at the Node level (fine), but there's no proxy-package-specific knob for users who'd rather not set a global env var system-wide.

End user impact: in a Zscaler/Netskope/Forcepoint/Bluecoat shop the proxy starts and answers `/health` 200 (loopback works), but every `/v1/messages` forward fails. The failure mode is opaque — looks like an Anthropic API outage from the user's perspective.

## What changed

| File | Change |
|------|--------|
| `package.json` | `dependencies.hpagent: ^1.2.0` (zero runtime deps; ~50 KB unpacked; supports both http and https upstreams in one package) |
| `proxy/config.mjs` | `envFlag` helper; new fields `httpProxy` (with both uppercase + lowercase env-var lookup), `noProxy`, `caFile`, `rejectUnauthorized` |
| `proxy/upstream.mjs` | Agent factory: `shouldBypassProxy()` (handles `*`, `.suffix.example`, bare host), `loadCa()` (graceful read failure → stderr + `undefined`), `getAgent()` with lazy init + one-time TLS-disabled stderr warning. `agent: getAgent(...)` wired into `options`. No-proxy/no-CA path returns `null`, preserving today's `agent: undefined` semantics. |
| `proxy/upstream.test.mjs` | New. 3 tests using `node --test`. |
| `README.md` | New "Corporate environments (proxies, custom CAs)" subsection under Quick Start: Proxy. |

## Defaults are preserved

- `getAgent()` returns `null` when no proxy / no CA / TLS-verify enabled → `transport.request({ ..., agent: null })` is treated by Node identically to the prior `agent: undefined` (uses the global default agent). No connection-pooling regression.
- Stderr is silent on the no-config path — the `[upstream] using proxy ...` line only prints when an agent is actually constructed.

## Tests

```
TAP version 13
ok 1 - forwardRequest routes through HTTPS_PROXY when set         (21ms)
ok 2 - forwardRequest bypasses proxy when host is in NO_PROXY     (4ms)
ok 3 - forwardRequest with no proxy env vars uses default agent (no regression)  (1ms)
1..3
# pass 3 / fail 0
```

The first test spins up a fake HTTP CONNECT proxy on a random loopback port, sets `HTTPS_PROXY` to it, and asserts that `forwardRequest` triggers a `CONNECT api.anthropic.com:443` against the fake. The second test sets `NO_PROXY: "api.anthropic.com"` and asserts the fake gets zero hits. The third test sets all proxy/CA env vars to empty and verifies the module loads + exports without error.

`node --test proxy/upstream.test.mjs` exits 0.

## Smoke (manual, on Windows 11 + Volta Node 22.22.0)

Without any env vars: proxy starts, `/health` returns 200, stderr clean. Behavior identical to v3.0.2.

With `HTTPS_PROXY=http://127.0.0.1:18888` against a tiny local mitm:
```
[upstream] using proxy http://127.0.0.1:18888 (rejectUnauthorized=true, ca=default)
```
mitm log shows `CONNECT api.anthropic.com:443`.

## Notes for review

- `hpagent` is the choice because it's zero-runtime-deps and supports both http/https upstreams in one import, which matters for a globally-installed CLI tool's install size. Happy to swap to `https-proxy-agent` if you prefer; behavior would be identical, just an extra dep.
- The bypass matcher follows the de facto `NO_PROXY` semantics used by curl/Python: bare host = exact match or `*.host` suffix; `.suffix.example` = subdomains only; `*` = bypass all.
- The tutorial-doc section 10 in the original write-up flagged `agent: false` as a footgun (disables keep-alive). I deliberately use `null`/`undefined`.
- `rejectUnauthorized: false` is only honored when explicitly opted into via `CACHE_FIX_PROXY_REJECT_UNAUTHORIZED=0`. A one-time `WARNING: TLS verification disabled` line is printed to stderr on first request so users can't accidentally ship insecure config.

## Cross-references

- [cnighswonger/claude-code-cache-fix-vscode#8](https://github.com/cnighswonger/claude-code-cache-fix-vscode/issues/8) — VS Code extension issue documenting corp-environment failure modes; that issue's accompanying VSIX patch already wires these env vars from VS Code settings into the proxy spawn env, but they're inert until this lands.
- [#52](https://github.com/cnighswonger/claude-code-cache-fix/issues/52) — prior Windows-pipeline-path fix from the same audit pass; merged as v3.0.2 (commit `aaa301e`). Same testing setup.

If any of this needs reshaping (split into multiple PRs, different proxy lib, different env-var names, different default for `rejectUnauthorized`), happy to revise.


cnighswonger:
Codex review:

Blocking finding:

- `proxy/config.mjs:27-28` collapses `HTTPS_PROXY` and `HTTP_PROXY` into a single `httpProxy` value chosen at module load, with `HTTPS_PROXY` taking precedence unconditionally. That means a plain-HTTP upstream (`CACHE_FIX_PROXY_UPSTREAM=http://...`) will still prefer `HTTPS_PROXY` when both vars are set, which is not standard proxy-env behavior and does not match the PR`s "Honor HTTPS_PROXY / NO_PROXY ..." framing. The current tests only exercise the default HTTPS upstream path, so this regression is not covered.

Non-blocking follow-up:

- `npm pack --dry-run` now includes `proxy/upstream.test.mjs` in the published tarball because `package.json` ships the whole `proxy/` directory. That is avoidable package bloat and publishes test-only code unnecessarily.

Verification I ran:
- `npm install`
- `node --test proxy/upstream.test.mjs`
- `npm test`
- `npm pack --dry-run`

Review artifact: `docs/code-reviews/pr-54-corp-proxy-ca-support-review-2026-04-23.md`

Bottom line: the feature is useful and mostly well-scoped, but I would not approve it yet because proxy-variable selection still appears wrong for plain-HTTP upstreams, and the new test file is being shipped in the npm artifact.

X-15:
Thanks for the detailed review and pushing the artifact straight to the branch — that made the loop tight.

Both blockers addressed in 9e51e9f + 2f5a8b5:

**Per-protocol proxy selection.** Split into `config.httpsProxy` and `config.httpProxy`, with `selectProxyUrl(isHTTPS)` matching curl/Python/Go semantics:
- https upstream → `HTTPS_PROXY`, falling back to `HTTP_PROXY`
- http upstream → `HTTP_PROXY` only (HTTPS_PROXY does NOT apply)

`selectProxyUrl` is now exported and table-tested with 8 cases covering all (HTTPS_PROXY, HTTP_PROXY) × (https, http) combinations + a separate test for lowercase variants.

**Test no longer ships in the npm tarball.** Moved `proxy/upstream.test.mjs` → `test/proxy-upstream-corp-proxy.test.mjs` (matching the existing `test/proxy-*.test.mjs` convention). `npm pack --dry-run` now omits it since `files` only ships `proxy/` and `bin/`.

**Side change worth calling out.** The new corp-proxy/CA fields on `config` are now getters so they reflect live `process.env` rather than the value at module init. Existing fields keep their prior read-once semantics. This made the unit testing tractable without spawning child processes per scenario.

**Side note on dropped integration tests.** I had two integration tests (HTTPS_PROXY routes through a fake CONNECT proxy / NO_PROXY bypasses it) that worked locally but kept the test process alive on Node 18 because hpagent's `keepAlive: true` sockets prevent clean exit (Node 20.4+ has `--test-force-exit`, 18 doesn't). I dropped those; hpagent's own test suite covers that path, and the protocol-based selection logic is now covered by direct unit tests on `selectProxyUrl`. Manual smoke is in the PR body. Happy to add them back via per-process spawning if you'd prefer end-to-end coverage in CI.

Also added a tiny CI fix in 9e51e9f: `npm install --no-audit --no-fund` before `npm test` since the workflow didn't install deps and the new `hpagent` import failed to resolve.

Pre-existing local Windows-only fail in `proxy server lifecycle / shuts down cleanly on SIGTERM` (verified against main without my changes too — it's a SIGTERM-on-Windows thing). Linux CI passes.

cnighswonger:
Codex review:

Re-review complete on commit `2f5a8b5` (artifact committed on top as `2a057d5`). The prior blocker is resolved:

- Proxy selection is now protocol-aware: HTTPS upstreams prefer `HTTPS_PROXY` and fall back to `HTTP_PROXY`, while HTTP upstreams use only `HTTP_PROXY`.
- The new `selectProxyUrl()` tests cover the protocol matrix, including lowercase env variants.
- `proxy/upstream.test.mjs` was moved out of `proxy/`, and `npm pack --dry-run` confirms the test file is no longer included in the published tarball.
- CI now installs dependencies before running tests, which is necessary after adding `hpagent`.

Verification I ran:
- `node --test test/proxy-upstream-corp-proxy.test.mjs`
- `npm test` (`198/198`)
- `npm pack --dry-run`

One non-blocking note: the dropped end-to-end fake CONNECT proxy tests mean this now relies on unit-level proxy selection coverage plus `hpagent` behavior for the actual tunnel path. That is acceptable for this scope, but a future child-process integration test would provide stronger coverage without keep-alive leaks.

Review artifact: `docs/code-reviews/pr-54-corp-proxy-ca-support-rereview-2026-04-23.md`

Bottom line: approved.

PR #77, opened 2026-04-25
feat: Docker image + GH Actions publish workflow (#49)

### LINKED ISSUE #49: Publish a docker image
Please publish a docker image to make it easier to run claude-code-cache-fix in a self-contained container.

cnighswonger:
Closes #49.

## What ships

- **`Dockerfile`** — `node:22-alpine` base, ~241 MB. COPYs only `package.json` + `proxy/` (proxy is the artifact; the install-service templates and other host-side tooling live outside the container). Runs as unprivileged `node` user (uid 1000). Binds `0.0.0.0:9801` (necessary for Docker `-p` forwarding; the publish layer gates external exposure). `HEALTHCHECK` directive runs `curl -fs /health` every 30s.
- **`.dockerignore`** — keeps build context small: excludes `.git`, tests, docs, preload, install-service templates, etc.
- **`.github/workflows/docker-publish.yml`** — publishes to `ghcr.io/cnighswonger/claude-code-cache-fix` on tag push (`v*.*.*`). Multi-arch (`linux/amd64` + `linux/arm64`) via Buildx + QEMU. Uses GHA cache for fast rebuilds. Semver-ladder tags: `3`, `3.1`, `3.1.1`, `latest`. `workflow_dispatch` input lets us rebuild ad-hoc with a custom tag for Dockerfile-only fixes without npm bumps.
- **`README.md`** — new "Docker" subsection in Quick Start: Proxy with run commands, env-var overrides, corp-CA mount example.

## Container behavior

```bash
docker run -d --name cache-fix-proxy --restart=always -p 9801:9801 \
  ghcr.io/cnighswonger/claude-code-cache-fix:latest
export ANTHROPIC_BASE_URL=http://127.0.0.1:9801
```

Auto-recovery via `--restart=always` (Docker-native; no systemd healthcheck companion needed in container land).

## Verification

Smoke-tested locally:
- Build: 5s on cached layers, ~6s cold
- Image size: 241 MB (alpine + node + ~hpagent dep)
- `docker run` + `curl /health` returns `{"status":"ok"}`
- Container logs: `proxy listening on 0.0.0.0:9801`
- `HEALTHCHECK` passes
- Container runs as uid 1000 (`node` user)

`npm test` on the branch: 465/465 green (no test changes; the Docker work is purely additive).

## Risk

Low. The workflow only fires on tag push (no inadvertent registry writes from regular PR merges to main). Multi-arch via QEMU adds CI runtime (a few minutes per release) but is standard. If the workflow itself has issues on first invocation, no published image is overwritten. The image is built from the same source the npm package is.

## First publish

This PR adds the workflow but doesn't publish — first published image will be on the next tag (e.g. v3.1.2 or v3.2.0). We can also kick a `workflow_dispatch` run after merge to publish a `:3.1.1` and `:latest` immediately if you want it before the next tag.

— AI Team Lead


cnighswonger:
Removing `approved-by-lead` and `ready-for-merge` pending human review.

Issue #14, opened 2026-04-13
feat: configurable TTL per request type (main + subagent)


cnighswonger:
## Context

Currently the interceptor forces `ttl: "1h"` on ALL outgoing cache_control blocks unconditionally. This was the right blunt-force fix when the main-turn TTL was broken pre-v2.1.90, but on v2.1.104+ where main turns are handled upstream, forcing 1h on subagent calls is arguably unnecessary overhead.

Additionally, the interceptor can be used beyond Claude Code — any Node.js application making Anthropic API calls can load it via `NODE_OPTIONS="--import"`. Pure API applications like weather simulation systems have different TTL needs than interactive CC sessions.

## Feature request (from @cowwoc on #10)

Make the TTL configurable per request type:

```bash
# Default (current behavior): force 1h on everything
# No env vars needed — backwards compatible

# Configure main-turn TTL (default: 1h)
export CACHE_FIX_TTL_MAIN=1h

# Configure subagent TTL (default: 1h, Anthropic's intended default is 5m)
export CACHE_FIX_TTL_SUBAGENT=5m

# Or for pure API apps that want full control:
export CACHE_FIX_TTL=5m          # set both to the same value
export CACHE_FIX_TTL_DISABLE=1   # don't inject TTL at all, let the caller decide
```

## Use cases

1. **CC users on v2.1.104+** — main turns handled upstream, want subagents back to 5m per Anthropic's intended design: `CACHE_FIX_TTL_SUBAGENT=5m`
2. **Pure API applications** — batch/one-shot workloads where 5m is cheaper and 1h is waste: `CACHE_FIX_TTL=5m`
3. **Long-running API sessions** — analysis pipelines that re-read cached prefixes across minutes: `CACHE_FIX_TTL=1h` (current default, no change)
4. **Users who want no TTL injection** — trust the caller's cache_control as-is: `CACHE_FIX_TTL_DISABLE=1`
5. **Intermediate values** — @cowwoc suggested 30m as a personal preference. API supports arbitrary durations? (needs verification)

## Detection: main vs subagent

The interceptor would need to distinguish main-turn calls from subagent calls to apply different TTLs. Possible heuristics:
- Message count (main turns accumulate messages; subagent first calls have few)
- System prompt content (subagents have different identity strings)
- The identity normalization feature (`CACHE_FIX_NORMALIZE_IDENTITY`) already detects Agent SDK vs CC — could reuse that detection

## Scope

[...]

PR #79, opened 2026-04-25
directive: overage cost warning extension (#47)

### LINKED ISSUE #47: Overage cost warning based on rate-limit warning headers
When the API returns `allowed_warning` status with `surpassed-threshold` and `upgrade-paths` headers (observed at Q7d=76%), the proxy should surface a user-friendly cost warning.

### What to show

- Current Q5h/Q7d utilization and projected time to 100%
- Estimated cost per hour at API overage rates based on actual burn rate
- Cost comparison: overage vs plan upgrade for the billing cycle

### Implementation

Extend `cache-telemetry.mjs` (or new extension) to:

1. Detect `allowed_warning` status and `surpassed-threshold` header in response
2. Accumulate token counts across requests (in-memory, resets on proxy restart)
3. Compute burn rate from accumulated tokens + utilization delta over time
4. Calculate projected overage cost at API rates
5. Emit one-time warning via stderr log + optional status line update

### Trigger

Only fires when these headers appear:
```
anthropic-ratelimit-unified-status: allowed_warning
anthropic-ratelimit-unified-7d-surpassed-threshold: 0.75
anthropic-ratelimit-unified-upgrade-paths: upgrade_plan,overage
```

### Target

v3.1.0 (alongside session serializer)

### Context

At our observed multi-agent burn rate, $100 of extra usage at API rates lasts ~3 hours. A plan upgrade from Max 5x to 20x ($100 for the billing cycle remainder) gives 4x headroom. The proxy is the right place to surface this because Anthropic's own headers suggest the action but don't provide the cost context.

cnighswonger:
## Stage: directive

Closes #47 (when implemented and merged).

This PR seeds the directive for a new opt-in proxy extension that surfaces actionable warnings when the user is approaching or has crossed Anthropic's overage threshold. The proxy already sees every response header — we can carry burn-rate state across calls and project hours-to-100% with the cost trade-offs spelled out, instead of the user discovering the overage on the next invoice.

- `docs/directives/proxy-overage-cost-warning.md` — full directive (goals, scope, detection rules, burn-rate math, dedup semantics, env vars, test plan, reviewer checklist).

### What this PR will contain after implementation
- `proxy/extensions/overage-warning.mjs` — the new extension (order 610, immediately after `cache-telemetry`).
- `proxy/rates.mjs` — extracted shared rate constants.
- `tests/overage-warning.test.mjs` — covers all 14 test-plan items.
- `extensions.json` entry (default `enabled: false`).
- README extension table entry.

1. **New extension, not a `cache-telemetry` extension.** `cache-telemetry` persists the latest snapshot; the overage warning is stateful (rolling window, dedup). Mixing them muddies the state model.
2. **In-memory state only for v3.2.0.** Proxy restart resets the dedup window. Documented limitation; file-backed dedup is more code than it's worth at this stage.
3. **Advisory only.** No request mutation. Stderr line + JSONL append. Status line integration is a downstream consumer concern.
4. **One-time-per-threshold-per-Q5h-window dedup.** Anthropic emits discrete `surpassed-threshold` values (0.75, 0.90, …) — used verbatim as the dedup key.

- Plan-upgrade decision UI (calculator)
- Cross-window Q7d projection (noisy over days)
- Persisting warning state across restarts
- Status line integration

Labeled `directive-stage` per the established workflow. Will move to `implementation-stage` once the directive is approved.

— AI Team Lead


cnighswonger:
Revised directive to address both Codex blockers and the three nits — pushed in `b9c2671`.

**Activation model** — switched from `enabled: false` + env var (which cannot work because disabled extensions are never loaded by the pipeline) to `enabled: true` in `extensions.json` with an internal `CACHE_FIX_OVERAGE_WARNING=1` env-gate that no-ops every hook on the first line when unset. Same pattern as `prefix-diff`.

**Hook lifecycle** — added an explicit per-hook responsibility table:
- `onResponseStart`: latch trigger eligibility into `ctx.meta._overageWarning` (header read only).
- `onStreamEvent` for `message_start`: append input/cache token sample to the rolling window.
- `onStreamEvent` for `message_delta`: append output sample, then if eligible-and-not-yet-emitted, compute projection + emit + flag emitted.

Single emission per response guaranteed by the `emitted` flag on `ctx.meta`. Cross-response dedup keyed by `(threshold, q5h_resets_at)` in module scope. Warm-up (<3 samples) emits the threshold-crossed warning with `null` projection fields and a fallback stderr line.

**Nits**
- Test path corrected: `test/` not `tests/`.
- Cost projection narrowed to a deliberately-coarse single weighted constant in `proxy/rates.mjs`, labeled `coarse` in stderr / JSONL field name (`cost_per_hr_usd_coarse`) / source comment. A precise per-tier cost engine is filed as a v3.3.0 follow-up.
- Concurrency contract now cites `fs.promises.appendFile` + Linux `O_APPEND` + `PIPE_BUF` (4096 bytes) atomicity directly, plus a 50-parallel-write integration test, instead of leaning on `usage-log` as precedent.

Ready for re-review.

— AI Team Lead

cnighswonger:
Pushing a small technical correction to test #14 — `PIPE_BUF` atomicity applies to pipes/FIFOs only, not regular files. Codex caught this on the equivalent claim in #80; correcting it here too even though this PR is already approved, since carrying an incorrect technical claim into implementation would be worse than amending the directive now.

The corrected wording is honest about what's actually guaranteed: POSIX guarantees `O_APPEND` offset+write atomicity (no torn offsets), but NOT non-interleaved writes on regular files. Linux ext4/xfs/btrfs empirically don't interleave single `write(2)` calls of <4 KB; test #14 validates that empirically; escalation paths (`tmp+rename`, `flock(2)`, single-writer queue) documented for when the empirical guarantee isn't enough.

Pushed in `0fe52fa`. No design changes; this is a tightening of the correctness claim.

— AI Team Lead

cnighswonger:
Both blockers fixed in `c2c5817`.

**Bug 1 — window not warming on non-trigger responses.** Split the sampling state from the emission gate:
- New `ctx.meta._overageQuota` is set by `onResponseStart` whenever the `q5h_util` header is present, regardless of whether this response is eligible to emit.
- `ctx.meta._overageWarning` continues to gate emission only.
- `message_start` now reads `_overageQuota` (not `_overageWarning.raw`), so every response with quota data warms the window.

**Bug 2 — `message_delta` leaking output_tokens to the prior response's sample.** Response-local sample handle:
- When `message_start` pushes a sample, it ALSO stores a reference at `ctx.meta._overageSample`.
- `message_delta` updates `ctx.meta._overageSample.output` directly — never `_window[_window.length - 1]`.
- A response that never sampled has no `_overageSample` and therefore cannot donate output to any other response's sample.

**New lifecycle tests** (L1–L3) — these are the cases the original tests didn't cover:
- L1: three non-trigger responses warm the window; a subsequent eligible response emits past warm-up with `window_samples ≥ 3`.
- L2: three `message_delta` events in one response → exactly one emission (the `emitted` flag is respected).
- L3: response A (no quota header → no sample) interleaved with response B (with sample) → A's `message_delta` cannot mutate B's sample. B's own `message_delta` still does (250 tokens land where they should).

21 tests pass, full suite 486/486.

Ready for re-review.

— AI Team Lead

Issue #18, opened 2026-04-14
quota-statusline.sh is not working


dmurat:
The quota-statusline.sh does not output anything. I believe this is because I do not have claude-meter.jsonl. What is claude-meter.jsonl, and how is it generated?

cnighswonger:
@dmurat Good catch — the status line was gated on `claude-meter.jsonl` with no fallback. Fixed: it now falls back to `quota-status.json` (written by the cache-fix interceptor) when claude-meter isn't installed.

Update to get the fix:
```bash
npm install -g claude-code-cache-fix@latest
```

Then re-copy the script:
```bash
cp "$(npm root -g)/claude-code-cache-fix/tools/quota-statusline.sh" ~/.claude/hooks/
```

`claude-meter.jsonl` is from a separate package ([claude-code-meter](https://github.com/cnighswonger/claude-code-meter)) — it's not required for the status line. The interceptor's `quota-status.json` has all the data the status line needs.

dmurat:
It works. Thank you.

PR #80, opened 2026-04-25
directive: upstream change detection extension (#39)

### LINKED ISSUE #39: Upstream change detection — alert on CC-originated cache-busting changes
## Problem

When Anthropic ships CC updates that change request structure, we discover the impact after users report cost spikes. Recent examples:

- v2.1.112: `cache_control` marker count changed from 2 to 3, breaking sticky (#35)
- v2.1.112: `[1m]` context tag silently removed from model string ([#50083](https://github.com/anthropics/claude-code/issues/50083))
- v2.1.113: Node.js replaced with Bun binary, killing preload interception (#35)

## Proposal

Add a detection layer that fingerprints the structural shape of each request and alerts when something changes unexpectedly. Detection targets:

- New/removed `system[]` blocks
- `cache_control` marker count changes
- Model string mutations
- Beta header additions/removals
- Tool count/schema drift
- New system-reminder patterns

This is read-only — no mutations, just detection and logging.

## Output channels

1. **Local debug log** — always available, writes to existing `cache-fix-debug.log`
2. **Dedicated alerts file** — structured JSON for programmatic consumption
3. **Remote telemetry endpoint** (opt-in) — sanitized structural fingerprints (block counts, marker positions, schema hashes — no prompt content, no user data) sent to a configurable remote endpoint. Enables:
   - Central alerting dashboard for enterprise/distributed deployments
   - Cross-installation early warning (one user's CC update triggers alerts for all)
   - Fleet-wide visibility across multi-host or multi-team environments

Remote telemetry would be off by default, enabled via env var (e.g. `CACHE_FIX_TELEMETRY_URL`).

## Implementation

Transport-agnostic — the detection logic works on the request body regardless of whether it runs in the preload or a future proxy. Designed as a standalone extension alongside the existing 16.

cc @deafsquad

cnighswonger:
## Stage: directive

Closes #39 (when implemented and merged).

This PR seeds the directive for a read-only structural fingerprinter that surfaces upstream CC changes proactively. Recent costly examples we discovered after-the-fact:

| Version | Change | How we found out |
|---|---|---|
| v2.1.112 | `cache_control` marker count went 2 → 3 | User cost-spike reports + manual diffing |
| v2.1.112 | `[1m]` context tag silently removed | CC issue #50083 |
| v2.1.113 | Bun binary replaced Node, killing preload | User reports + retro investigation |

A passive structural fingerprint compared call-to-call would have surfaced each within minutes of the first invocation.

- `docs/directives/proxy-upstream-change-detection.md` — full directive (fingerprint shape, diff/alert semantics, baseline persistence, env vars, test plan, reviewer checklist).

### What this PR will contain after implementation
- `proxy/extensions/upstream-change-detection.mjs` — order 50 (early, before any mutating extension)
- `tests/upstream-change-detection.test.mjs` — 16 cases covering stability, sensitivity, persistence, throttle
- `extensions.json` entry (default `enabled: false`)
- README extension table entry

1. **Fingerprint contains NO prompt content.** Only counts, positions, bucketed sizes, and hashes of immutable identifiers (model, beta headers, tool names). A reviewer should be able to read the fingerprint shape and confirm zero user data leaks even if the JSONL file is shared.
2. **Per-namespace baselines** keyed by `(model_string, beta_headers_set)` — different models or beta-header combos get separate baselines so a `claude-haiku-*` call doesn't trip an alert on the `claude-opus-*` baseline.
3. **First-ever fingerprint = `baseline_established` event, not an alert.** Subsequent diffs become alerts. A new baseline replaces the old one atomically; we don't keep alerting on the same change.
4. **Throttle: 10 alerts in 60s → 1 `alert_storm` event + 5min suppression.** Belt-and-suspenders against a fingerprint bug producing infinite alerts.
5. **Baseline persists across proxy restarts** via `~/.claude/upstream-baseline.json` (atomic write, tmp + rename, unique tmp suffix). No false-positive flood on restart.

- Remote telemetry endpoint (`CACHE_FIX_TELEMETRY_URL`) — deferred to v3.3.0+ pending privacy review
- Cross-installation correlation dashboard
- Auto-remediation
- Classifying changes as cache-busting vs benign — v3.2.0 reports the diff verbatim; humans decide

Labeled `directive-stage`. Will move to `implementation-stage` once the directive is approved.

— AI Team Lead


cnighswonger:
Revised directive to address all four Codex blockers and the four nits — pushed in `0c34f0d`.

**Activation model** — switched from `enabled: false` + env var to `enabled: true` + `CACHE_FIX_UPSTREAM_DETECTION=1` env-gate (same fix as #79).

**Content-free fingerprint** — this was the most substantive rework. Replaced:
- `block_role_markers` (parsed section names like `"Environment"`) → `known_section_marker_set_hash` (sha256 of sorted allowlist indices) + `known_section_marker_count` + `unknown_section_marker_present` boolean.
- `system_reminder_patterns_seen` (extracted tag text like `["<system-reminder>", ...]`) → `known_reminder_pattern_set_hash` + `known_reminder_pattern_count` + `unknown_reminder_pattern_present` boolean.

The "unknown present" booleans use tightly-bounded shape regexes (`/^# [A-Z][a-zA-Z ]{1,30}$/`, `/<[a-z][a-z-]{1,30}>/`) that record ONLY whether matches existed, never the matched text. When the boolean flips false → true, that's the signal "Anthropic shipped a new tag we haven't seen" — investigation requires opting into a separate (not-in-v3.2.0) one-shot diagnostic dump.

Added **test #18**: fingerprint computed from a body containing `"SECRET-TOKEN-XYZ"` must satisfy `JSON.stringify(fingerprint).includes("SECRET-TOKEN-XYZ") === false`. Unit-level enforcement of the content-free invariant.

**Fingerprint equality** — removed `timestamp` from the fingerprint payload entirely. Timestamps now live in event records only (`baseline_established`, `structural_change`). Equality is a straight structural comparison.

**JSONL atomicity** — explicit contract section. `fs.promises.appendFile(path, recordString + "\n")` as a single syscall, relying on POSIX `O_APPEND` + `PIPE_BUF` (4096 bytes on Linux) atomicity. Each event record is well under 4 KB. Documented the upgrade path (per-record segmentation) if records ever grow past that bound. Test #14 spawns 50 parallel writes and asserts no truncation or interleaving.

**Nits**
- Hook lifecycle table added (module load / `onRequest` / hot reload — explicit state ownership).
- **Throttle removed from scope.** Codex was right — adding another time-window state machine to a feature whose fingerprint design already prevents alert storms is overkill. Rationale recorded in the Out-of-Scope section: an alert fires only when fingerprint changes, and the new baseline replaces the old one immediately, so the same change cannot fire twice. If a fingerprint bug ever did produce an alert storm, that's a bug to fix at the source, not to suppress at a throttle.
- Test path corrected: `test/` not `tests/`.
- Atomic-write test re-spec'd: inject a `rename` failure (mock fs), assert prior baseline file intact, assert tmp file cleaned up in `finally`. Replaces the brittle "tmp file collision" test.

Ready for re-review.

— AI Team Lead

cnighswonger:
Fixed the remaining blocker — pushed in `d6083a1`.

You're right: `PIPE_BUF` atomicity applies to pipes/FIFOs only, not regular files. The previous draft overclaimed. The corrected contract is honest about what's actually guaranteed:

- POSIX guarantees the `O_APPEND` open atomically combines offset adjustment with the write — concurrent appenders don't lose data to torn offsets.
- POSIX does NOT guarantee non-interleaved writes to regular files at sub-write granularity. That's a kernel/filesystem implementation detail.
- In practice on Linux with ext4/xfs/btrfs and a single `write(2)` of a buffer well under one page, record-level interleaving does not occur. **Empirical kernel behavior, not POSIX.**
- Test #14 (50 parallel writes, assert no interleaving) is the empirical validation. If it ever fails on a supported platform, the implementation must escalate.

The directive now spells out why we accept the empirical guarantee for v3.2.0 (small records, rare events, diagnostic log not transactional state) and documents the upgrade path (per-record `tmp+rename`, `flock(2)`, or single-writer queue) when the tradeoffs change.

Note: the same incorrect `PIPE_BUF` claim appears in #79 (which you previously approved) and #81. I'm pushing the same correction to both even though #79 is already approved — better to fix the wrong technical claim now than carry it into implementation.

Ready for re-review.

— AI Team Lead

cnighswonger:
Both blockers fixed in `cea8f7d`.

**Bug 1 — beta header source**: `extractBetaHeaders` now reads the `anthropic-beta` request header (with case-insensitive lookup) as the primary source, and falls back to `body.anthropic_beta` only as a defensive measure. `computeFingerprint(body, headers)` and the `onRequest` hook were both updated to thread `ctx.headers` through. New test **9b** asserts that flipping the header from empty → present actually changes the namespace hash — that test would have failed before this fix.

**Bug 2 — prefix-substring marker matching**: replaced the includes-based search with strict line-based matching. The text is split on `\n` into a Set, and each allowlist entry is checked for exact equality against the set. `# Environment Details` no longer matches the allowlist index for `# Environment`. New tests **17c** and **17d** assert no prefix-substring leakage AND that exact-line matches still work even when a similar-but-longer line appears elsewhere in the same text.

26 tests, all green. Full suite 512/512.

Ready for re-review.

— AI Team Lead

Issue #11, opened 2026-04-12
fix: stabilize git-status in system prompt to prevent cache busting


cnighswonger:
## Context

@wadabum reported on anthropics/claude-code#47107 that `includeGitInstructions` (default: true) injects live `git status` output into the `system[]` prompt. Every file edit, stage, or unstage changes the git status, which changes the system prompt, which invalidates the entire prefix cache.

This is a cache-busting mechanism UPSTREAM of the TTL tier — it fires regardless of whether you're on 5m or 1h TTL because the prefix itself changes.

## Proposed fix

Stabilize the git-status portion of the system prompt before the request goes out, similar to how we stabilize the fingerprint. Options:

1. **Strip git-status from system prompt entirely** — most aggressive, may lose useful context
2. **Hash or normalize the git-status output** — replace the volatile portion with a stable placeholder while preserving the instructional text around it
3. **Move git-status to a later position** in the system blocks so it doesn't invalidate the prefix of everything before it
4. **Env-gated opt-out** — `CACHE_FIX_STABILIZE_GIT_STATUS=1` to strip/normalize, off by default until validated

Option 4 is safest for a first pass. The git-status text lives in the last `system[].text` block per wadabum's analysis.

## References

- anthropics/claude-code#47107 — original report
- anthropics/claude-code#45756 — wadabum's comment linking this to the quota discussion
- anthropics/claude-code#47098 — companion issue about skills/CLAUDE.md blocks in messages[0] being uncacheable

wadabum:
Hi, 
-> bypass in the form of `env.CLAUDE_CODE_DISABLE_GIT_INSTRUCTIONS=1` or setting [includeGitInstructions=false](https://code.claude.com/docs/en/settings#available-settings)
were you able to cross-validate this?




On 'fix'
**3. likely should go to the very-very end**... 
if not **1. totally remove it** (and rely on function calling if it wants a `git status`? - maybe a hint to this)
or even *new* **5. pack it into the first user-message** ((_after_ system-reminder´s / projects-claude.md))
heck even *new* **6. after the first user message** may be reasonable...

Think scenario: run `claude "30k tokens about how to review the latest changes"` - switch branch, `rerun 'your analyzer query'` 
-> could be full hit, instead those 30k will always miss ((and `--append-system-prompt` isnt rly a fix for that either))

**_rly neither of those paths would "generally" work well..._**
[...]

PR #81, opened 2026-04-25
directive: claude-meter compat with proxy mode (#70)

### LINKED ISSUE #70: Make claude-code-meter compatible with proxy mode
## Background

`claude-code-meter` (separate npm package, repo: `cnighswonger/claude-code-meter`) was wired into the legacy CC wrapper via `NODE_OPTIONS="--import .../claude-meter/preload.mjs"`. With v2.1.113+ shipping CC as a Bun binary, `NODE_OPTIONS` no longer applies — the meter can't load on modern installs. The README references it as an opt-in sharing mechanism, which is now silently broken for any user on Bun.

We've removed those references in #N (incoming PR — link when filed) so the README stops promising a path that doesn't work. This issue tracks the actual fix: make claude-meter compatible with proxy mode.

## Proposed approach

The proxy already produces structured per-call telemetry via the `usage-log` extension (`~/.claude/usage.jsonl`) and per-window quota state via `cache-telemetry` (`~/.claude/quota-status.json`). The cleanest integration:

1. **Enable `usage-log` by default** (currently `enabled: false` in `extensions.json`). Or document it as the opt-in for users who want claude-meter ingestion.
2. **Refactor `claude-code-meter`** to be a downstream consumer of `~/.claude/usage.jsonl` rather than an in-process preload. It can run as a separate background process, a cron-driven uploader, or a manual `claude-meter upload` command.
3. **Remove the wrapper hack** — `~/bin/claude` no longer needs the legacy NODE_OPTIONS load path.

This makes claude-meter:
- Work for both Bun and Node CC installs
- Decouple from CC binary lifecycle
- Reuse the proxy's already-vetted telemetry (no duplicate collection logic)

## Out of scope for this issue

- The data-sharing protocol itself (consent, anonymization, upload endpoints) — already designed in claude-code-meter
- Whether to ship `usage-log` enabled by default in v3.1.0 (separate decision)

## Related

- Wrapper code: `~/bin/claude` lines 35-36 (legacy load path)
- Proxy extension: `proxy/extensions/usage-log.mjs`
- Removed README references: tracked in incoming PR
- Future direction note in TRACKED_ISSUES.md (NEXO Brain row mentions a `claude-meter → NEXO` exporter)

— AI Team Lead

cnighswonger:
## Stage: directive

Closes #70 (when implemented and merged).

This PR seeds the directive for the cross-repo refactor that restores `claude-code-meter` for users on CC v2.1.113+ (Bun binary, NODE_OPTIONS no longer applies). The fix: expand the proxy's existing `usage-log` extension to a superset record schema matching claude-meter's `MeterRowSchema`, then refactor claude-meter to ingest from `~/.claude/usage.jsonl` instead of via the broken preload.

- `docs/directives/proxy-claude-meter-compat.md` — full directive (cross-repo scope split, schema reconciliation, v:2 superset shape, test plans for both sides, reviewer checklist).

### What this PR will contain after implementation (cache-fix side only)
- `proxy/extensions/usage-log.mjs` expanded to v:2 schema with sessionId, q5h/q7d delta tracking, all 24+ MeterRowSchema fields
- `tests/usage-log.test.mjs` — 13 cases per test plan
- README updates documenting the new schema and the claude-meter integration path
- `extensions.json` retains `enabled: false` default (default-on is a v3.3.0 decision)

### What lands in the claude-meter repo (separate PR there)
- New `src/ingest/jsonl-tailer.mjs` — reads usage.jsonl forward from a saved offset
- `bin/claude-meter.mjs` adds `ingest [--source <path>] [--once] [--watch]` subcommand
- `src/log/schema.mjs` accepts v:2; marks proxy-not-yet-emitting fields optional
- `src/interceptor/preload.mjs` gets a deprecation notice; `package.json` drops the `./preload` export

1. **Single canonical wire format.** The proxy's `usage.jsonl` becomes the source of truth. claude-meter validates each row through its existing `MeterRowSchema`. No duplicate collection logic.
2. **Schema-version 2.** Bumping `v: 1 → 2` is an explicit breaking change for anyone parsing the old 9-field shape. Documented in CHANGELOG.
3. **Optional vs required fields tracked exactly to MeterRowSchema constraints.** Optional fields (`org_id`, `qoverage_util`, `qrepresentative_claim`, `overage_disabled_reason`) emitted only when source headers are present; previously-required-with-default fields (`web_search_requests`) emitted as `0` when absent.
4. **`org_id` is hashed with sha256 prefix, never raw.** Same privacy guarantee claude-meter has always had.
5. **`enabled: false` stays as the default for v3.2.0.** Users opt in. We can revisit default-on in v3.3.0 once adoption is observed and disk-usage concerns surface.

- Default-enabling `usage-log` (deferred v3.3.0)
- Share-protocol or upload-endpoint changes (claude-meter side unchanged)
- Backfill of historical preload-written `~/.claude/claude-meter.jsonl`
- Real-time streaming beyond watch-mode polling

Labeled `directive-stage`. Will move to `implementation-stage` once the directive is approved.

— AI Team Lead

### DIFF
--- a/docs/code-reviews/pr-81-claude-meter-compat-directive-review-2026-04-25.md
+++ b/docs/code-reviews/pr-81-claude-meter-compat-directive-review-2026-04-25.md
@@ -1,0 +1,37 @@
+# Review: claude-meter compat directive
+
+Date: 2026-04-25
+Reviewed: `docs/directives/proxy-claude-meter-compat.md`
+Label applied: `changes-requested`
+
+## What Is Correct
+- The high-level direction is sound: moving claude-meter off the broken `NODE_OPTIONS` preload and onto proxy-produced telemetry is the right fix for CC v2.1.113+.
+- Reusing the existing `usage-log` extension pattern is the right activation model for this feature. Keeping the extension opt-in via extension config, rather than introducing a second env-var gate inside the extension, is consistent with the repo’s extension loader and with the current `usage-log` design in `proxy/extensions/usage-log.mjs`.
+- Most of the raw data needed by claude-meter is available on the proxy side. The request model is already captured in `proxy/server.mjs`, rate-limit headers are available in `responseHeaders`, and the SSE stream includes the usage payloads that carry `speed`, `service_tier`, `cache_creation.ephemeral_*`, and `server_tool_use.web_search_requests`.
+- Splitting the work across the two repos is directionally correct: this repo should produce the canonical per-call log row, and claude-meter should switch ingestion sources without changing its upload/share protocol.
+
+## Blockers
+- The directive’s “canonical superset” schema is not actually compatible with claude-meter’s current strict validator. `MeterRowSchema` is a `z.strictObject` with `v: 1` and does not allow extra keys such as `total_input` or `peak_hour`, while the directive says the proxy will emit `v: 2` plus those extra fields and that claude-meter will “reuse existing `src/log/schema.mjs` validation.” Those statements cannot all be true at once. The directive needs one coherent contract: either the shared wire schema is updated in claude-meter to accept the new fields/version, or the proxy output must stay within the strict row shape that claude-meter validates today.
+- The activation/config story is wrong in two places. The directive tells users to enable the integration with `CACHE_FIX_USAGE_LOG=1`, but in this repo `CACHE_FIX_USAGE_LOG` is the path override, not an enable flag, and `usage-log` is supposed to remain a config-toggle opt-in. It also says to add a comment to `extensions.json`, but this repo parses that file with `JSON.parse`, so comments are not valid there. The directive needs to describe the real activation path: extension config enables the module; `CACHE_FIX_USAGE_LOG` only overrides the destination path.
+- The org ID privacy requirement is not specified precisely enough for a cross-repo contract. The directive says “hashed sha256 prefix” and has a test that asserts “not the original,” but it never states the exact encoding and truncation rule. claude-meter’s existing writer uses `sha256(...).digest("hex").slice(0, 16)`. If the proxy side is supposed to preserve compatibility and privacy guarantees, the directive needs to pin that exact algorithm and output length.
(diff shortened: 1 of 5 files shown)


cnighswonger:
Substantial rewrite to address all six Codex blockers — pushed in `efcde34`.

**The biggest change: abandoned the v:2 superset approach.** Codex was right that emitting v:2 with extra fields like `peak_hour` against `z.strictObject({ v: z.literal(1), ... })` cannot work. The new contract: **proxy emits EXACTLY MeterRowSchema v:1**. No new schema version. `peak_hour` dropped from the proxy JSONL output (trivially recomputable from `ts` if any consumer needs it). The schema file in `claude-code-meter` becomes the single source of truth.

**Activation corrected** to match the actual `usage-log` pattern, not the misremembered one:
- Default `enabled: false` lives in the extension's own export.
- Users opt in by adding `"usage-log": { "enabled": true, "order": 650 }` to `proxy/extensions.json`.
- `CACHE_FIX_USAGE_LOG=<path>` is path override only, NOT an enable flag.
- No env var for activation. No comments in `extensions.json` (JSON.parse).

**`org_id` privacy pinned exactly** to claude-meter's algorithm: `sha256(raw).digest("hex").slice(0, 16)`. Specified in the field-source table, the implementation choice section, the test plan (test #11 asserts bit-exact match), and the reviewer checklist.

**State capture across stream events** now explicit: `message_start` populates `ctx.meta._usageLog` with model/speed/service_tier/ephemeral split/web_search_requests; `message_delta` reads from it and adds `output_tokens` to assemble the final record. Tests #4 and #5 cover this directly.

**Cross-repo release ordering** spelled out as a hard requirement: proxy ships first (v3.2.0), then claude-meter v0.4.0 declares `>= cache-fix 3.2.0` as the supported producer. The two PRs are NOT independently shippable. Stated in both repos' CHANGELOGs.

**Migration story**: old 9-field rows in users' existing `~/.claude/usage.jsonl` files fail claude-meter's strict validator and are skipped on the reader side with a debug log. No backfill. Documented in CHANGELOG as a one-time event.

**Test path corrected** to `test/proxy-usage-log.test.mjs` (the existing file — we expand it, don't create a new one).

Also added: a full 29-field source table covering every MeterRowSchema field with its exact origin (request body, message_start, message_delta, response header, derived). Expanded reviewer checklist to 16 items including the "v field is the literal number 1" check, the "peak_hour absent" check, and the bit-exact org_id hashing check.

Ready for re-review.

— AI Team Lead

cnighswonger:
Pushing the same `PIPE_BUF` correction Codex caught on #80 — applies here too. Implementation choice section's atomicity claim now reflects what's actually guaranteed: POSIX `O_APPEND` offset+write atomicity (real), NOT non-interleaved regular-file writes (Linux empirical, not POSIX). Test #15 (50 parallel writes, no interleaving) is the empirical validation. Escalation paths documented.

Pushed in `efb9789`. No design changes; tightening of the correctness claim.

— AI Team Lead

PR #84, opened 2026-04-27
feat: image-strip — add CACHE_FIX_IMAGE_MAX_DIM oversized-image guard

cnighswonger:
## Why

Asked for by Alexander (X-15) and useful for Chris's Chronology hi-res scan workflow. Prevents the Anthropic API error:

> "An image in the conversation exceeds the dimension limit for many-image requests (2000px). Start a new session with fewer images."

When the user attaches a large image (5K-10K px scan, photo, screenshot at retina res), this error fails the request entirely. Until now the only mitigation was to manually pre-resize before attaching.

## What's in the PR

- **New env var `CACHE_FIX_IMAGE_MAX_DIM=<pixels>`**. Default unset = no enforcement (current behavior unchanged). Set e.g. `2000` to enable.
- **New module `proxy/image-dimensions.mjs`** — pure-JS PNG and JPEG header parsing, no native deps.
- **`image-strip` extension extended**: scans ALL images (not just OLD ones — orthogonal to KEEP_LAST), strips any whose width OR height exceeds the limit. Covers BOTH user-message direct images AND tool_result-nested images.
- **Forensic placeholder** preserves original dimensions: `[image stripped — exceeded 2000px max dimension (was 3000x1500px)]`
- **Composes with KEEP_LAST**: keep_last runs first (drops oldest), max_dim then runs on what remains. Both stat objects persist on separate `ctx.meta` fields for back-compat (`imageStripStats` shape unchanged; new `imageStripOversizedStats` for the new pass).
- **Fail-open**: images whose dimensions can't be parsed are kept rather than stripped. Better to send a request that might error than to strip a valid image we just couldn't read.

## Tests

- 16 new in `test/proxy-image-dimensions.test.mjs` (synthesized PNG + JPEG headers, dispatch, malformed inputs, case-insensitive media_type)
- 11 new in `test/proxy-image-strip.test.mjs` (boundary cases at exactly the limit, both image locations, KEEP_LAST + MAX_DIM composition, fail-open)
- All 36 image-related tests pass; **full suite 553/553**

## Out of scope (explicit, deferred to v3.3.0+ if requested)

- Image RESIZING — would need `sharp` or similar native dep. Strip-only is the no-deps minimum viable fix.
- GIF/WebP/AVIF/BMP detection — not in CC's typical pipeline.

## Activation

Opt-in. The `image-strip` extension itself is `enabled: false` by default in `extensions.json` — users explicitly turn it on. Then set `CACHE_FIX_IMAGE_MAX_DIM=2000` (or other ceiling) to enable the new oversized guard. Both env vars compose if desired.

## Review request

Codex first via `cli_delegate`, then Alexander (since he asked for the feature). Holding for both reviews + Chris's read before any merge or release decision.

— AI Team Lead

### DIFF
--- a/README.md
+++ b/README.md
@@ -340,9 +340,20 @@ Keeps images in the last 3 user messages, replaces older ones with a text placeh
 export CACHE_FIX_IMAGE_MAX_DIM=2000
 ```
 
-Anthropic enforces a per-image dimension ceiling on multi-image requests; any image exceeding the limit (currently 2000px on a side) fails the entire request with `"An image in the conversation exceeds the dimension limit for many-image requests (2000px). Start a new session with fewer images."` Common triggers: hi-res scans, retina screenshots, photos at full resolution.
+The Anthropic API enforces TWO image-related limits on multi-image requests, and the same error message can fire for either:
 
-`CACHE_FIX_IMAGE_MAX_DIM=2000` enables a per-request scan that strips oversized PNG/JPEG images (in both user messages and tool results) and replaces each with a forensic placeholder noting the original dimensions. Pure-JS header parsing — no native deps. Composes with `CACHE_FIX_IMAGE_KEEP_LAST` (keep_last runs first, max_dim then runs on what remains). Other formats (GIF, WebP, AVIF, BMP) pass through unchanged.
+> `"An image in the conversation exceeds the dimension limit for many-image requests (2000px). Start a new session with fewer images."`
+
+Two pressure axes to address them:
+
+| Pressure | Variable | What it does |
+|---|---|---|
+| **Too many images in conversation** | `CACHE_FIX_IMAGE_KEEP_LAST=N` | Strips images from old user messages, keeps only the last N. |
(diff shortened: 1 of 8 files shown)


cnighswonger:
@X-15 — this is the dimension-cap feature you asked about. Codex reviewed and approved (`approved-by-codex-agent`). Tested with PNG and JPEG synthetic headers; composes with `CACHE_FIX_IMAGE_KEEP_LAST` if you want both running. Fail-open on unparseable headers — better to send a request that might error than to strip a valid image we just couldn't measure. Strip-only by design (no native deps); resize would need `sharp` and that's a Phase 2 if anyone asks. Default off — opt in with `CACHE_FIX_IMAGE_MAX_DIM=2000`.

Whenever you have time, would appreciate a second set of eyes — both for whether the env-var ergonomics match what you had in mind, and for the case where it should/shouldn't fire that I might not have anticipated.

— AI Team Lead

X-15:
When the user attaches a large image (5K-10K px scan, photo, screenshot at retina res), this error fails the request entirely. Until now the only mitigation was to manually pre-resize before attaching.
--------
It also appears when too many images are in the conversation

PR #155, opened 2026-05-28
statusline: round bar tick like fill

schuay:
draw_bar rounded fill but truncated the tick. With consumed=15% and elapsed=19.6% that gave fill=2 and tick=1, placing the tick inside the filled run ([█┃█░░░░░░░]) — the visual reserved for over-pace. Field- reported on a clearly under-pace Q7d.

Round both: monotonicity keeps tick at-or-past fill whenever consumed <= elapsed. T17 locks the reported case; T8 shifts by one cell because the symmetric round is more accurate.

Fyi this is the last of my planned fixes on this.

### DIFF
--- a/docs/code-reviews/pr155-statusline-bar-tick-codex-review-2026-05-28.md
+++ b/docs/code-reviews/pr155-statusline-bar-tick-codex-review-2026-05-28.md
@@ -1,0 +1,23 @@
+# Review: statusline bar-tick rounding
+
+Date: 2026-05-28
+Reviewed: PR #155 (`tools/quota-statusline.sh`, `test/quota-statusline-smoke.test.mjs`)
+Label applied: changes-requested
+
+## What Is Correct
+- The patch targets the reported regression directly: T17 reproduces the 15% consumed / 19.64% elapsed case and the shared cell conversion moves the tick back to the fill boundary.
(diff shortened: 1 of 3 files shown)


schuay:
> Fyi this is the last of my planned fixes on this.

Never mind, I'm already chasing the next one (fixing burn rate calc to not depend on assumed window start). I don't know; maybe I'm overengineering this and we should just keep it simple, even if it sometimes results in obviously wrong output like `Q7d [░░░░░░░░┃░] 1% (exhaust 533d20h, reset 1d14h)` - after an unscheduled quota reset.

> now relies on Python's banker's round() for both fill and tick, which collapses some real over-pace cases back to the boundary.

That's intentional due to the low bar resolution. We *could* artificially move the tick one index to the left/right of the fill to display small over/underconsumption. But IMO it's cleaner to honestly show the tick and fill at the same position if they're at similar levels.

cnighswonger (approved):
Going with this as-is.

We'll open a follow-on issue for the "burn-rate depending on assumed window start, producing [░░░░░░░░┃░] 1% (exhaust 533d20h, reset 1d14h) after an unscheduled reset" question just to keep track of it.

Many thanks @schuay for your work on this! It has been super-helpful in my own sessions to have the visual reference.

Chris

PR #190, opened 2026-06-04
Implement server.mjs debug logging

nisqatsi:


cnighswonger (approved):
A human was here... 👀

PR #246, opened 2026-07-02
Honor CLAUDE_CONFIG_DIR for all on-disk proxy state

codeslake:
First off, thank you for this project. We've been running it daily and it's saved us a lot of quota. Really appreciate the work. Small contribution back.

## Problem

The proxy hardcodes `os.homedir()/.claude` for every runtime path it writes: `quota-status/`, `usage.jsonl`, `cache-fix-state/`, session mirrors, snapshots, OAuth events, and the debug log. Claude Code, however, reads `CLAUDE_CONFIG_DIR` to relocate its config root away from the default `~/.claude`.

When you run **one proxy per config dir** (each pointed at a different `CLAUDE_CONFIG_DIR`), every proxy still writes to `~/.claude/quota-status/account.json` and they clobber each other's quota state on every response, so statuslines read the wrong session's cache stats and `account.json` bounces between values. This is the same class of problem #104 fixed for per-session files, one level up (per config dir).

## Fix

Add a small helper `proxy/claude-home.mjs`:

```js
export function claudeHome() {
  return process.env.CLAUDE_CONFIG_DIR || join(homedir(), ".claude");
}
```

and route all 15 runtime path builders through it instead of the hardcoded `join(homedir(), ".claude", ...)`.

- When `CLAUDE_CONFIG_DIR` is **unset**, `claudeHome()` returns `~/.claude`, i.e. **identical behavior to today**, so single-config users see no change.
- When it's **set**, each proxy's state lands under its own `$CLAUDE_CONFIG_DIR`, so proxies pointed at different config dirs stay cleanly separated.

The helper reads the env live (not cached at module load) to match the existing test-isolation pattern in `config.mjs`.

## Scope

- `proxy/claude-home.mjs` (new, 12 lines)
- 15 files: mechanical `homedir()/.claude` -> `claudeHome()` substitution across `proxy/server.mjs`, `proxy/session-mirror-writer.mjs`, `proxy/oauth/*`, and the `proxy/extensions/*` that write state.
- README: a short `CLAUDE_CONFIG_DIR` subsection under "Why per-session".

No behavior change when `CLAUDE_CONFIG_DIR` is unset. `node --check` is clean on all touched files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -655,6 +655,12 @@ The shipped [`tools/quota-statusline.sh`](tools/quota-statusline.sh) is the refe
 
 On multi-agent hosts (multiple Claude Code sessions sharing one proxy), the pre-v3.5.0 single global file caused every session to overwrite the others' cache stats with each response. A statusline reading from session A would show session B's TTL tier whenever B sent a request more recently. Per-session files plus an account-global quota file resolve this without losing the easy account-wide view. See [#104](https://github.com/cnighswonger/claude-code-cache-fix/issues/104) for the original report.
 
+### `CLAUDE_CONFIG_DIR`
+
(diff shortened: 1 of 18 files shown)


cnighswonger (approved):

cnighswonger:
@codeslake Thanks for this fix!

Issue #248, opened 2026-07-02
Optional forward-proxy mode to keep Remote Control working (ANTHROPIC_BASE_URL disables it on CC >= 2.1.196)


codeslake:
Thanks again for the project. Raising this as an issue first to check appetite before sending a larger PR.

## Problem

The proxy is wired in via `ANTHROPIC_BASE_URL`. On Claude Code **>= 2.1.196**, pointing `ANTHROPIC_BASE_URL` at a non-Anthropic host **disables Remote Control** (`/remote-control`), `/schedule`, and claude.ai MCP connectors. CC treats any custom base URL like a Bedrock/Vertex gateway and gates those features off. So today it is cache-fix **or** Remote Control, not both.

(The binary confirms the gate: "Remote Control is only available when using Claude via api.anthropic.com.")

## Idea

Add an opt-in **forward-proxy mode** (`CACHE_FIX_FORWARD_PROXY=on`) where the proxy sits in front of the real `api.anthropic.com` as an `HTTPS_PROXY` instead of as the base URL:

- The proxy also handles HTTP `CONNECT`. It MITMs **only** the upstream host (terminating TLS with a locally-generated CA so the existing extension pipeline runs unchanged) and **blind-tunnels every other CONNECT** (mcp-proxy, telemetry, etc.) untouched.
- The client sets `HTTPS_PROXY` + `NODE_EXTRA_CA_CERTS` and leaves `ANTHROPIC_BASE_URL` **unset**, so CC's base URL stays `api.anthropic.com` and Remote Control keeps working.
- Non-transformed paths (RC credential fetch, OAuth, etc.) are relayed to upstream verbatim; only `/v1/messages` runs the pipeline.

Reverse-proxy mode stays the default and is unchanged; this is purely additive behind a flag.

## Status

I have a working implementation running daily: forward-proxy core + passthrough for non-transformed paths + a WebSocket-upgrade relay (so `/voice` keeps working) + a lock around the generated-CA step for concurrent starts. It also plugs into `install-service` (systemd/launchd) and the Docker image so forward-proxy is a first-class option across all launch methods, matching how `ANTHROPIC_BASE_URL` works today.

Mainly want to check this fits how you think about the interception point before sending a big diff. Would you be interested in a PR for this?

codeslake:
For reference, I pushed the working implementation to a branch on my fork (not opening it as a PR yet, pending your thoughts on the approach):

[...]

PR #251, opened 2026-07-11
Add opt-in forward-proxy mode to keep Remote Control (implements #248)

codeslake:
## Summary

Adds an **opt-in forward-proxy mode** so the cache-fix proxy keeps prompt-cache savings **without** disabling Claude Code's Remote Control.

Implements the approach proposed in #248.

## Problem

The default (reverse-proxy) mode sets `ANTHROPIC_BASE_URL` to point at the local proxy. On Claude Code **≥ 2.1.196**, any non-`api.anthropic.com` base URL trips the Remote Control gate, so RC is disabled (and on ≥ 2.1.201 the `/rc` command is hidden from the palette entirely). Users who want cache savings currently have to give up Remote Control.

## Approach

Run the proxy as an **`HTTPS_PROXY`** endpoint instead of a base-URL override:

- CC's `ANTHROPIC_BASE_URL` stays **unset** → base URL remains `api.anthropic.com`, so the RC gate never trips.
- The proxy handles `CONNECT` and **selectively MITMs only `api.anthropic.com`** (where the cacheable `/v1/messages` traffic lives), using a locally generated CA. Every other `CONNECT` is **blind-tunneled** untouched.
- Non-`/v1/messages` paths on the MITM'd host (e.g. `/api/features/`) are passed straight through to the real backend, so feature flags and auth behave exactly as they would without the proxy.
- WebSocket `Upgrade` on the MITM'd host is relayed to real upstream.

Net: `/v1/messages` is still transformed for cache savings, and everything RC depends on reaches the genuine Anthropic backend.

## Opt-in

Off by default. Enable with:

```
CACHE_FIX_FORWARD_PROXY=on
```

Nothing changes for existing reverse-proxy users unless they set that.

## Client wiring

```
export HTTPS_PROXY=http://127.0.0.1:<port>          # default 9801, or $CACHE_FIX_PROXY_PORT
export NODE_EXTRA_CA_CERTS=~/.claude/cache-fix-ca/ca.pem   # or $CACHE_FIX_CA_DIR/ca.pem
# leave ANTHROPIC_BASE_URL unset
```

`install-service` accepts `CACHE_FIX_FORWARD_PROXY=on` so the systemd unit / launchd plist / Docker container run the proxy end supervised; the README's new "Running it persistently" subsection documents the three client-wiring options. The service manages the proxy process only — the client still wires `HTTPS_PROXY` + `NODE_EXTRA_CA_CERTS` itself.

## Changes (2 commits, 9 files, +468/-5)

- `proxy/forward-proxy.mjs` (new, 248 LOC) — CONNECT handler, selective MITM, blind-tunnel, CA gen/self-heal, WS relay
- `proxy/server.mjs` — wire `attachForwardProxy()`, `handlePassthrough()`, `/health` reports `forward_proxy`
- `proxy/config.mjs` — `CACHE_FIX_PROXY_PORT`, `forwardProxy` getter, version in `/health`
- `bin/claude-via-proxy.mjs`, `bin/install-service.mjs` — forward-proxy launch + supervised install
- `templates/*.service.template`, `templates/*.plist.template`, `Dockerfile` — forward-proxy line/env
- `README.md` — forward-proxy section + supervised-running docs

## Testing

- Boots on a throwaway port: `/health` → `{"status":"ok","forward_proxy":true}`, CA auto-generated (`ca.pem`/`ca.key`/`leaf.pem`/`leaf.key`).
- `node --check` clean on all changed `.mjs`.
- Verified end-to-end on **Claude Code 2.1.207**: Remote Control stays enabled through the proxy, and cache savings apply on resumed sessions.
- Passes the newer (2.1.201+) `tengu_ccr_bridge` feature-flag gate as well, because `/api/features/` is passed through to the real backend rather than rewritten.

## Scope / non-goals

- Reverse-proxy mode is unchanged and remains the default.
- Does **not** modify `ANTHROPIC_BASE_URL`, `/v1/messages` semantics beyond the existing cache transform, or any managed `disableRemoteControl` setting.

## Note on #246

Independent of #246 (Honor `CLAUDE_CONFIG_DIR`). This branch is rebased onto `main` and carries **only** the forward-proxy change — no overlap with #246's config-dir commit. The two can merge in either order.

---


cnighswonger:
Smoking this locally this afternoon. I should get it merged later today.

codeslake:
Thanks for the thorough pass — the live end-to-end validation and the Codex review both landed, and all three blockers were real. Pushed fixes to `feat/forward-proxy` (head `8bafd63`); each blocker has a dedicated commit with a targeted regression test.

**Blocker 1 — CA-gen lock race + unconditional lock deletion** (`050e94b`)
Generation now requires positive lock ownership. `acquire()` stamps the owner pid inside `.gen.lock`; a waiter that times out reclaims the lock only if the owner is dead (`ESRCH` / no pid stamp after the full wait), and throws instead of generating while a live owner still holds it. The `finally` removes the lock only when this process created it. Temp files are now per-pid and cleaned up, so two starters can't clobber each other's `.tmp.*` (this also clears the `.tmp.leaf.csr`/`.tmp.leaf.ext` litter tracked as #253 item 1). The 30s wait is `config.caLockWaitMs` so the contention paths run in ms under test.

**Blocker 2 — leaf publish not atomic across key+cert** (`03d7fd6`)
`ready()` now proves the on-disk `leaf.key` matches `leaf.pem` (SPKI public-key compare) on top of existence + SAN, so a mixed-generation key/cert pair is treated as not-ready and regenerated instead of served. Closes the window regardless of the two-rename publish order.

**Blocker 3 — attach-failure fallback + unremoved self-heal handlers** (`81b5430`)
Routing and `/health` now key on `_forwardActive` (a count of successfully-attached forward instances), not `config.forwardProxy`, so a failed attach serves reverse-mode 404s instead of relaying non-`/v1/messages` paths upstream. The `uncaughtException`/`unhandledRejection` swallowers install only on a successful attach, are ref-counted across instances (replacing the one-way `__CACHE_FIX_SELF_HEAL_ON` env guard), and are removed on `close()` — an embedded/shared process regains Node's default crash semantics once the forward instance is gone.

**Non-blocking, folded in** (`8bafd63`): `ca.key`/`leaf.key` normalized to `0600` on both generation and reuse; the stale "global (outside CLAUDE_CONFIG_DIR)" comment corrected.

New coverage:
- `test/proxy-forward-ca.test.mjs` — live-owner lock is respected (no generate, lock untouched); dead-owner and pid-less stale locks are reclaimed and cleaned up; 8 concurrent starters all end with a key/cert pair chaining to the single CA, no lock or temp litter left; a mismatched `leaf.key` is rejected and regenerated; key modes are normalized.
- `test/proxy-forward-attach-fallback.test.mjs` — attach failure (scrubbed PATH → `openssl` ENOENT) keeps 404s with zero upstream relays and no swallowers; a successful attach installs them, `close()` removes them, and a later reverse-only instance in the same process stays reverse throughout.

Full `node --test` suite is green. Left the remaining lighter items on #253. Thanks again — glad the forward-proxy shape held up against a live session.

cnighswonger (approved):

PR #272, opened 2026-07-29
feat(insertion-normalization): pin volatile reminder blocks so mid-history rewrites stop busting the cache

Gunther-Schulz:
## The bust class

Claude Code re-serializes old `<system-reminder>` hook blocks later in the session — moving one into its own message, or merging it into a neighbour. That edits a message mid-history, so everything after the edit index is re-billed. Reported independently with the same diagnosis in anthropics/claude-code#76606; on our capture corpus it shows up as the `splice/insert-mid` class, ~40 kB re-billed per unmitigated hit, several times per evening session.

## The mitigation

`insertion-normalization` keeps a per-conversation canonical model of the history, keyed by **content identity** (`message-hash.mjs`: content hash + occurrence ordinal — position-independent, and repeated identical reminders stay distinct; a history that carried the same reminder 44 times taught us why). Volatile blocks are pinned to their first-seen serialization: when CC later re-shapes an old reminder, the forwarded bytes keep the canonical form and the cached prefix survives. A history that stops matching the model (compaction, genuine rewrites) resets honestly instead of forcing a stale canon — pins survive the reset, order assumptions do not.

- Off by default: `CACHE_FIX_INSERTION_NORMALIZE=1` enables normalization, `CACHE_FIX_VOLATILE_PIN=1` the pinning.
- State persists under the state dir; proxy restarts are transparent to it.
- `message-hash.mjs` is shared identity infrastructure — a follow-up PR (deferred-tool-rewrite) builds on it.

## Evidence

Run in production on this fork for a full working day of heavy sessions (~2.5 GB of pre/post captures). Every observed `splice/insert-mid` pair forwarded with 0 re-billed bytes; cross-request stability, sequence and canonical-order invariants all 0 violations under a daily replay gate. The 46-test suite includes bite tests for the reset/pin coupling and the occurrence-ordinal identity, each demonstrated red on the real defect before the fix.


Gunther-Schulz:
Production soak status, for merge timing — stated so the evidence base is explicit rather than implied:

- In production on our machines since 2026-07-28 (gates on). The newest part is the reset/pin coupling fix (pins survive an honest reset) — ~1 day old, found by the replay gate in live traffic and bite-tested.
- A scheduled daily gate re-runs this exact code over our full capture corpus (currently ~2.5 GB of real multi-day traffic) under the four invariants plus a fidelity check anchored to the recorded forwarded-body hashes. Latest run: 9 captures, 0 violations.
- First full day with the complete stack live (2026-07-29): zero preventable cold rewrites — the only cache-cold event was a 0-context overnight TTL expiry. The prior day, while these fixes were still landing incrementally, logged 15 cold rewrites / ~5.2M re-billed tokens on the same workload profile. One day, uncontrolled workload — a signal, not a measurement.

We'll post a week-of-soak summary on this thread as the daily gate accumulates runs. Happy to hold the merge until then if you prefer.

Gunther-Schulz:
One consequence worth knowing during review: the canonical content identity here supersedes `messages-cache-breakpoint` — its #47098 head-mutation rationale is covered by recognising a message across re-serializations instead of pinning index 0. On our fork we removed that extension after this landed, measured clean across the corpus (0 stability/safety/sequence). The removal is already up as #281 (draft, stacked on this branch so the replacement is present in it) — opened now rather than after merge so the plan is visible during this review; it carries the default-behavior trade-off, since this PR's gates are off by default while `messages-cache-breakpoint` ships enabled. Whether to take it at all is your call. Sequencing note from the measurement: sequencing note from the measurement: if you also run any mid-history breakpoint scheme, remove that first — freeing the 4th breakpoint slot *arms* it (measured 57 self-inflicted busts in that order).

Gunther-Schulz:
Pushed c713d0e: the pin now also suppresses the migrated duplicate (anthropics/claude-code#76606). Background: CC can migrate a hook reminder out of its tool_result into a standalone system message mid-history; pinning restored the first-seen form but the migrated copy still forwarded, so the same content went upstream twice — measured live as a ~61 kB mid-history splice re-billing 124k tokens on one turn. A standalone message whose wrapper-normalized bytes equal a live pinned block is now suppressed (never forwarded, never given canonical identity); genuine changes still forward and reset. New test file covers the helpers, the classify-level behavior, cross-request stability, and the genuine-change escape (57 pass / 1 skip in this slice — the real-capture red-green check needs the replay tooling and runs in #276, where it passes against the original capture).

Gunther-Schulz:
Pushed wave 2 for this slice — five commits. 1ca82f0 adds the real-pair suppression test with a pinned-fixture fallback (the 15k-line harvested capture rides along, sanitized, 0 addresses); a270da0 handles the merge variant — a standalone reminder equal to the join of its pinned blocks is suppressed (the shape behind a measured 587k-token re-bill); a80c481 stops tail-stripping — a final-message duplicate is payload, never stripped. 47ff9c3 makes the test file slice-portable: the fixture reader now loads dynamically like the replay tools, so in this tree the real-capture check reaches its designed skip instead of failing at module load; it executes in #276 where the tooling lives. b713b2f carries the oscillation fixture the merge test reads. Tests here: 67 pass / 1 skip (the designed one) / 0 fail; byte-equal to fork tip da9bf8c across every shared path.

Gunther-Schulz:
Agreed on the fixture path, and thank you for the "don't push a fix
commit yet" catch — you're right that a scrub-on-top leaves the
original blobs reachable. Plan, for your confirmation before anything
lands:

1. **We rewrite this branch** so the identifying fixture blob never
   becomes reachable from your `main` — the original fixture never
   existed in the rewritten history; only a fully sanitized
   replacement (new UUID-free filename) does. Force-push once you
   confirm.

2. **The sanitizer itself gets the fix first, on our fork.** While
   preparing follow-up branches we found two more gaps in the same
   scrubber, so the rewrite will carry a strictly stronger
   sanitization than a spot-fix would: (a) image blocks passed through
   raw — the scrubber redacted `block.data` but wire images nest
   base64 at `block.source.data`, one level down; (b) exactly the
   structural-identifier class you flagged: conversation keys,
   per-request `sid`s, wall-clock timestamps, and the session UUID in
   the filename. New scheme: content bytes and images one-way
   tokenized (`t_<sha12>_<len>` per paragraph, preserving the `"\n\n"`
   join relations the tests assert; `data_<sha10>` for binary), keys
   and sids replaced by deterministic sha-derived tokens, timestamps
   rebased to a fixed epoch keeping only intra-fixture deltas,
   filenames carrying the token instead of the UUID. Every rebuilt
   fixture is replay-verified to reproduce identical classifier
   verdicts, and a mechanical test asserts the absence classes (no
   raw base64 runs, no live timestamps, no UUIDs) so the sanitization
   claim is checked, not just stated.

On the other blockers, all accepted:

- **Blocker 2 (reminder-only byte change re-served stale):** agreed
  this is the load-bearing safety question and not patchable in the
  diff — we'll write the directive you asked for before the next
  review round. It will lead with a measurement: how often pinned
  reminder bytes actually change across matched entries in our
  capture corpus. If the answer is effectively never, the directive
  argues the evidenced allowlist; if it happens, the design is
  fail-closed re-pin (store the new bytes, honest reset of that
  boundary only). Either way the evidence rides in the directive.
- **Blocker 3 (state files at ambient umask):** agreed, and agreed
  it's a series-wide pattern — we'll fix it once as a pattern
  (explicit owner-only modes on every conversation-derived write,
  hashes instead of raw bytes where bytes aren't structurally
  required) rather than per-PR.
- **Blocker 4 (read-dedupe adjacency assertion):** will be the
  deliberate call you asked for — we'll check whether the adjacency
  is load-bearing for read-dedupe before deciding between moving the
  order and updating the assertion, and put the reasoning in the
  commit.

Nothing further lands on this branch until you've confirmed the
rewrite path in (1).

— Claude (drafted for and approved by Gunther Schulz)

Gunther-Schulz:
The directive requested before the next review pass is ready: [insertion-normalization — identity scheme, pin safety, persistence, and fixture strategy](https://github.com/Gunther-Schulz/claude-code-cache-fix/blob/40c11b2da78f8e6008dc29b2fb9ae753df70b1c2/docs/directives/insertion-normalization-identity-directive.md) (fork main, `40c11b2`; NFR section included, Load-bearing: **yes**).

Headlines, with the full measurement in `docs/code-reviews/blocker2-volatile-change-measurement.md` (same tree):

- **Blocker 2 answered by measurement, not argument.** Across 36 captures / 196 conversations / 11,074 same-conversation pairs, first-seen comparison per pinned identity — a strict superset of what the live pin compares (it can over-report change, never under-report): **IN-PLACE-TEXT = 0**. No reminder's text was ever replaced inside a still-matching message. The zero is informative, not vacuous — the corpus carries 119 distinct reminder texts across 10 kinds, so in-place variation is possible by construction; it lives across messages, never inside a settled one. All observed change is removal — the exact class the pin exists to absorb.
- **Decision: evidenced allowlist, monitored — not fail-closed re-pin.** The census now computes IN-PLACE-TEXT on every daily gate sweep, so a first real occurrence surfaces mechanically; that occurrence is the build trigger for the already-designed fallback (fail-closed re-pin: store the new bytes, honest reset of that boundary only). Failure modes are enumerated per class in the directive.
- **Persistence settled once for the series** — owner-only 0600 for all conversation-derived writes, raw bytes only where structurally required (`entry.m` re-serving), hashes otherwise. Carries to #275/#280 unchanged.
- **Fixture strategy:** synthesized by default; harvested-and-scrubbed only as a per-fixture justified exception, committable only with the mechanical absence scan green — now enforced at test time and again at the push boundary via a pre-push scan. The 432 KB fixture gets a ≥10× cut with a verdict-identity acceptance check (replayed classifier verdicts identical pre/post cut).

The rewrite package for this branch is ready as coordinated (original fixture blobs never reachable from your `main`; sanitizer-hardened, minimized fixtures only) — still holding until you confirm the rewrite path; nothing force-pushes before that. Stacked PRs rebase in order afterward (#273 → #276 → #278, #281 last).

— Claude (drafted for and approved by Gunther Schulz)

Gunther-Schulz:
The rewrite is done, per the path you confirmed — force-pushed, history replaced.

**New head:** 7 commits on current `upstream/main` (`23346ac`), tip `720ecb4`. Verified at the object level before pushing: `git rev-list --objects` over the whole branch against `upstream/main` reaches none of the pre-sanitization fixture blobs, under any name; a per-commit identifier scan over every slice-changed file is clean at all 7 commits; and no fixture↔capture pairing comments remain. The fixtures the branch carries (`pinned-…`, `oscillation-…`) enter history already token-named and fully scrubbed.

One thing the scan caught beyond the fixture blobs, fixed in the same rewrite: a hardcoded live capture id + home path had entered with the branch's second commit (a test's real-capture default). Both stages now recover the capture by hashing candidate filenames against the fixture's own token — no identifier in source at any commit.

**Suite at the new head:** 1550 pass / 1 fail / 1 designed skip — the failure is the known `proxy-read-dedupe` extension-order adjacency assertion already on this thread.

**Stack status:** #276 got the same treatment (history rewritten, rebased, force-pushed — details on its thread, including the sanitizer hardening + in-tree absence scan it now ships). #281 is rebased onto this head. #273 and #278 are blob-clean and will rebase in your stated order (#273 → #276 → #278, #281 last) as the review proceeds. The join-move / reserved-entry work is up as draft #295, stacked on this branch and #276 — it closes the identity mis-binding where one extra copy of a recurring reminder re-binds an entry CC has stopped sending; corpus A/B: stability violations 10 → 2, both survivors belonging to `deferred-tool-rewrite` and identical in both trees.

Ready for the fresh round from the top whenever suits.

Gunther-Schulz:
Pushed `c489f29` — the five comment strings you named are scrubbed, and the grep turned up six more.

**The five you named** now carry the fixture's synthetic token: `insertion-normalization.mjs:616,659`, `test/insertion-merge-suppression.test.mjs:2`, `test/insertion-suppression.test.mjs:7,267`.

**Six you did not**, in the same class — comment-only measurement notes referencing three *other* real captures: `insertion-normalization.mjs:219,721,758,808,915` and `test/insertion-normalization.test.mjs:938`. These got a different treatment on purpose: the identifier is dropped and the shape description kept, rather than reused as the fixture token, because reusing it would attribute those measurements to the wrong capture. That is the fallback you suggested yourself.

Worth naming why your list undercounted, because it is a property of the check rather than of your reading: a diff-scoped read reaches what the slice changed, and these sat in lines the slice never touched. A corpus-wide grep for the bare `s-<8hex>` shape is what found them.

Verification, real output:

- `grep -rnE 's-[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}'` over the branch → 0 hits.
- `grep -rnE 's-[0-9a-f]{8}\b'` → 0 real capture ids. Three incidental matches remain and are not identifiers: `claude-3-opus-20240229` version strings in `tools/rates.json`, `tools/update-rates.mjs`, `tools/cost-report.mjs`, where `s-20240229` matches the shape by coincidence.
- `npm test` → 1567 tests, 1565 pass, 1 skipped, **1 fail**.

About that one failure, stated plainly rather than waved past: it is `test/proxy-read-dedupe.test.mjs:505`, the extension-order adjacency assertion — your blocker 4 from round 1, still open as a design question. I did not take it on trust that it was pre-existing: I checked out the branch base `720ecb4` into a scratch worktree and ran that suite there, and it fails identically at the base, before any of this scrub. The diff touches no extension ordering, no `extensions.json`, and not that test file — every changed line in this push is a comment.

Everything here is comment-only, so no behaviour changes. None of the eleven occurrences was a fixture key or a test assertion; had one been, it would have come back to you as a question rather than being rewritten.

Gunther-Schulz:
Rebased onto current `main` (`39570db`) — force-pushed as `60a9e3c`. Picks up #310 as you described.

**Verified rather than assumed, since a rebase re-parents eight commits:**

- Test 28 now reads `read-dedupe loads at order 380, after image-retry-circuit-breaker (370) and before cache-control-normalize (400)` and asserts `reg[idx].order === 380`. It passes with `insertion-normalization` at 395. `c15c136` is in this branch's ancestry.
- `test/proxy-read-dedupe.test.mjs`: 42 pass, 0 fail.
- Full suite in the worktree: **1708 pass, 0 fail, 1 skipped** (1709 tests, 56 suites, 25.0 s).
- All eight of our commits are patch-identical to their pre-rebase counterparts (`git patch-id --stable`, one-to-one), so nothing about the diff you approved has changed — only the base.

No code change on our side, as you said. Ready for a fresh CI run.

One note for whoever lands this, since it will recur on every rebase of an already-open branch: our push-side leak scan blocked the force-push on four `capture-key-prefix` findings, all of them in the commit messages of `0fc8cea`'s and `23dcc91`'s pre-existing history. Those messages are byte-identical to ones already on the remote, so they are already in `refs/pull/272/head` and the push could not add exposure — the scan ranges over the ref update, which after a rebase re-flags history it has already seen. Bypassed deliberately for that reason; the real fix is scoping the scan to content new to the remote, which is now booked on our side.

cnighswonger (approved):

PR #273, opened 2026-07-29
feat(deferred-tool-rewrite): hold tools[] byte-stable; announce additions via the mid-conversation beta

Gunther-Schulz:
> **Stacked on #272** — includes its commit (`message-hash.mjs` identity + the insertion-normalization system-prompt sub-key). Review the tip commit; will rebase once #272 lands.

## The bust class

Every ToolSearch/deferred-tool load makes Claude Code re-send a different `tools[]` array. `tools[]` heads the cache prefix, so each load re-bills the whole context. This is the class in anthropics/claude-code#81967 (there triggered by LSP add/remove — the deferred-tool path hits it far more often; at 700k context one load re-bills ~700k tokens).

The remarkable part: **the API already ships the fix.** The documented `mid-conversation-tool-changes-2026-07-01` beta lets a request announce an added tool in a `tool_addition` block at the *tail* of the conversation, leaving the prefix untouched — and Claude Code 2.1.220 carries that beta's documentation inside its own binary without using it on the wire.

## The mitigation

- `tools[]` is frozen at its first-seen form per session (keyed session + system-prompt + conversation, so subagents/sidecars never share a baseline — the keying collision was a measured bug before it was a design note).
- A mid-session addition keeps the frozen bytes; the new tool is announced via a `tool_addition` system message anchored at the tail, re-injected at a stable anchor on every subsequent request, beta header added. Removals/reorders are held byte-stable; a schema change resets honestly.
- **Opt-in per model, evidence required.** The beta is rolled out per model family; an unsupported model rejects the whole request with a 400 (measured: `claude-sonnet-5`, `claude-haiku-4-5` — haiku's error names the gating capability, "requires a model that supports mid-conversation system content"). Unknown models degrade to the status-quo bust, never a lost request, and the first suppressed announcement per model logs a warning naming the way out.
- `tools/probe-tool-addition.mjs` measures a candidate model in one real request; `CACHE_FIX_TOOL_ADDITION_EXTRA` admits a candidate on a throwaway proxy for a live probe. Allowlisted with wire evidence: `claude-opus-5`, `claude-fable-5` (capture holds the injected block; replaying the pipeline reproduces the forwarded body hash byte-for-byte; outcome record shows the API streamed a 200).

Off by default: `CACHE_FIX_TOOL_REWRITE=1`.

## Evidence

[...]

PR #275, opened 2026-07-29
feat(capture): pre-pipeline request capture + source/gate provenance on /health

Gunther-Schulz:
> **Stacked on #274** (header forwarding fix — this branch includes its commit). Review the tip commit.

## Why

This is the attribution primitive behind #272/#273: a JSONL capture of what Claude Code **actually sent**, recorded before any mutating extension runs (order 60). A divergence present in the raw capture is CC's; one absent is ours. Without this line there is no way to distinguish a mitigated upstream bug from a self-inflicted one — we spent a full day learning that the expensive way (five of six suspected CC bugs turned out to be our own, found only by replaying raw captures).

It's also what makes independent verification possible for *anyone running this proxy*: with captures on disk, the replay/census tooling (next PR) can re-run the whole pipeline offline against real traffic and prove — not assume — that the mitigations forward what production forwarded.

## What

Three record types share one capture file per session:

- **request** — pre-pipeline body + the headers that matter, joined by id;
- **outcome** — what the API charged (usage split by cache tier, output tokens, wall time) plus `outSha`, the hash of the bytes the proxy *actually forwarded*, computed at the single point those bytes exist (post-pipeline, in `preForward`). This is what lets an offline replay validate itself against production byte-for-byte;
- **boot** — restart boundaries and the gate set in force, so cross-restart analysis doesn't misattribute.

`/health` grows two provenance fields on the same principle (answer content questions, not label questions):

- `proxy_tree` — content fingerprint of the source **this process loaded** (mtimes false-fire on byte-identical restores; disk diverges from the process the moment someone edits without restarting);
- `gates` — the `CACHE_FIX_*` environment this process is actually serving with. The unit file answers "declared", not "serving" — that difference cost us a day of green gate runs that were verifying a pipeline nobody ran.

Size-capped (`CACHE_FIX_CAPTURE_MAX_MB`, oldest-file rotation). Off by default: `CACHE_FIX_REQUEST_CAPTURE=1`. Captures record header names relevant to caching only — no authorization material is written.

## Evidence

15 unit tests (record shapes, join ids, rotation, fingerprint stability across byte-identical restores). In production on our machines for two days at ~2.5 GB/day of captures; the outcome-record `outSha` is what proved the replay faithful (109/109 reconstructions matched the wire on a fresh session) and what proved fable-5's `tool_addition` support in #273.

### DIFF
--- a/proxy/extensions.json
+++ b/proxy/extensions.json
@@ -1,5 +1,6 @@
 {
   "bootstrap-defense": { "enabled": true, "order": 45 },
+  "request-capture": { "enabled": true, "order": 60 },
   "ttl-tier-detect": { "enabled": true, "order": 75 },
   "cc-version-normalize": { "enabled": true, "order": 90 },
   "fingerprint-strip": { "enabled": true, "order": 100 },
--- a/proxy/extensions/request-capture.mjs
+++ b/proxy/extensions/request-capture.mjs
@@ -1,0 +1,319 @@
+// request-capture — record full request bodies for offline replay.
+//
+// Directive: docs/directives/proxy-request-capture-replay.md (stage 1).
+// The proxy is the only component that sees every request byte-for-byte;
+// until this extension, it threw the bodies away, so every pipeline
+// change could only be validated against synthetic fixtures or live
+// traffic. Captures feed tools/replay.mjs and tools/cache-sim.mjs.
+//
+// Order 60 — after bootstrap-defense (45) and ttl-tier-detect (75 is
+// AFTER, fine: it only reads), before cc-version-normalize (90), the
+// first extension that MUTATES the body. Capture must record what CC
+// sent, not what the pipeline made of it.
+//
+// Activation: `enabled: true` in extensions.json (always loaded), runtime
+// gate CACHE_FIX_REQUEST_CAPTURE=1 (read per-call so tests can flip it).
+// Fail-open: capture failure never fails the request. CACHE_FIX_DEBUG=1
+// logs swallowed errors, same idiom as prefix-diff.
+//
+// Sensitivity: captures contain the FULL conversation content — the same
+// sensitivity as the transcripts in ~/.claude/projects/, on the same
+// machine. No new exposure class; documented in the directive.
+//
+// Retention: CACHE_FIX_CAPTURE_MAX_MB (default 2048). Checked every
+// SWEEP_EVERY appends per process; oldest capture files deleted first
+// until under the cap. The sweep is best-effort and fail-open.
+
+import { appendFile, mkdir, readdir, stat, unlink } from "node:fs/promises";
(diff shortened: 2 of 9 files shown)


Gunther-Schulz:
Pushed `8a47da4` (force-with-lease after the rebase) — all three parts of the round.

**Rebase.** Onto current `upstream/main`. The header-forwarding commit this branch carried has landed upstream in the meantime and was dropped as already-applied, so the stack is gone and this is now a **single commit** on top of `upstream/main`. The only conflict was one import line where both sides had added to it; resolved by keeping both. Nothing was cherry-picked away.

**Permissions.** Capture directory 0700, capture files 0600, applied at CREATE rather than chmod'ed afterwards — so there is no window in which a file holding whole request bodies exists at the ambient umask. The directory matters on its own account: its listing leaks session keys through filenames. Writes go through the repo's existing write-owner-only helper rather than a second implementation, which also gives the migration case for free — a file created before the helper existed is repaired on the next append, and captures are append-only and long-lived.

**The `/health` env dump.** This one turned out worse than the review framed it, so the fix is shaped differently than "an allowlist of gate keys".

Of the ~117 `CACHE_FIX_*` names this codebase reads, the switches are inert to publish while others carry an OAuth client id, a token endpoint, a credentials path, the upstream URL, a command line, and a dozen filesystem paths describing the operator's machine. A dump cannot tell them apart, and `/health` answers anything that can reach the port.

So: an allowlist of the gates whose **value** answers "which gates was this proxy running with", and **name-only for everything else**. The direction is the point — a redaction list covers the hazards someone thought to enumerate and ships the unanticipated one by default, whereas a name-only default makes a variable added tomorrow safe on the day it is added, without anyone remembering `proxy/gate-allowlist.mjs` exists. The reader still learns the variable is set, which is all provenance needed. The boot record uses the same function.

Measured cost of the change: **zero**. Fed the production gate set from the systemd unit, all eleven keys keep their values — nothing diagnostic was traded away.

Verification, real output:

- `npm test` → **1565/1565**, green.
- Red-first, and the arrangement is worth stating because a weaker one was available and would have looked identical in a report. Helper-level assertions pass with the call sites reverted — which is exactly the regression that matters — so the suite also drives `ext.onRequest` end to end and stats what actually landed on disk. With the two new modules present but the call sites reverted, precisely those two integration tests fail and the helper-level ones correctly do not.
- One test pins the **rule** for adding to the allowlist rather than its current membership: no key naming a path, URL, command or credential. It earned its keep immediately by failing an over-broad first draft of my own, where `*_LOG` (an on/off switch here) was conflated with `*_LOG_PATH` (a location).

Noted from your side of the thread: you have this marked load-bearing, so your human review follows. Nothing here needs chasing from us.

cnighswonger (approved):

PR #278, opened 2026-07-29
feat(output-guard): last-line invariant on the response path — restore, log, never corrupt

Gunther-Schulz:
> **Stacked on #272** (imports `validateToolAdjacency`). The `test/fixtures/replay-classes/` files are shared with #276 — identical content, merges cleanly in either order.

## Why

Once a proxy runs body-mutating extensions (#272/#273, or any future one), the failure mode that matters most isn't a missed optimization — it's shipping a *corrupted response*. This guard is the last line: it runs after every mutator and asks a different question than any unit test can — is what we're about to send still the response upstream sent, in every way the client depends on?

## What

- `output-guard-stash` (order 55) stashes the pristine body before any mutator runs.
- `output-guard` (order 690) validates structural invariants on the outbound response (tool_use/tool_result adjacency among them). On violation it **restores the stashed original** and logs the diff — a mutation bug becomes one logged line instead of a broken session.
- Gated off by default: `CACHE_FIX_OUTPUT_GUARD=1`.

## Evidence

10 tests, including gate 1: the guard fires **zero** times across every class-matrix corpus (a guard that fires on legitimate traffic trains its reader to ignore it — that's a failing guard, and its first 243 live firings on our fleet were exactly that: its own missing exemption, since fixed and pinned). Running in production alongside #272/#273 for two days, 0 safety violations across ~2.5 GB of daily replay.


Gunther-Schulz:
Pushed wave 2 for this slice — three commits. ed43a46 extends output-guard with the assistant-terminal invariant: the forwarded tail must stay assistant-terminal-equivalent to what came in; a mutator that strips the trailing message is caught. Two sync commits bring insertion-normalization and its suppression test to the #272 tip. As designed, the suppression test's real-capture check skips in this slice (no pinned fixture here — it executes in #276); a portability fix now makes that skip actually reachable, where previously the file failed at module load. Tests: 74 pass / 1 skip (the designed one) / 0 fail; byte-equal to fork tip da9bf8c across the four shared paths.

Gunther-Schulz:
Rebased onto current `main` (`b00b141`) — `3c4ecfa`, force-pushed.

**The review target shrank, and by more than the rebase alone.** Both sync commits are gone: `58cd931` touched `insertion-normalization.mjs` and `insertion-suppression.test.mjs`, both of which ship on main now via #272 — and git dropped `e4bd379` on its own with `patch contents already upstream`. So `insertion-suppression.test.mjs` (the 464 lines you'd counted) is no longer part of this PR's delta; it is main's file, unmodified by us. `git diff upstream/main HEAD -- proxy/extensions/insertion-normalization.mjs test/insertion-suppression.test.mjs` is empty.

What's left is 13 files, +543:

```
 proxy/extensions.json                          |   1 +
 proxy/extensions/output-guard.mjs              | 201 +
 proxy/extensions/output-guard-stash.mjs        |  32 +
 test/output-guard.test.mjs                     | 286 +
 test/fixtures/replay-classes/*.jsonl           |  8 files, 20 +
```

**One conflict needed a real decision**, in `proxy/extensions.json`: main now carries `request-capture` at order 60 (yours, via #275) on the line our `output-guard-stash` at 55 wanted. Both kept, ordered by their `order` value — 45 / **55** / 60 / 75. No semantic interaction; the stash simply has to precede the capture, and it does.

**Suite on the rebased branch**: 1724 tests, 1723 pass, 0 fail, 1 skipped (24.4s).

On your structural note about orders 55 and 690 — agreed, and worth recording that the risk is asymmetric: a future extension at >690 defeats the guard *silently*, which is the failure mode that earns an assertion rather than a comment. We have it filed on our side as an implementation follow-up for if/when the pattern lands; happy to fold it into this PR instead if you'd rather it ship with the guard than after it.

https://claude.ai/code/session_01Vz8u7MRdkZbCZ7RTbq5zMo

Gunther-Schulz:
Rebased onto current `main` (`e3149ae`) — the second rebase you asked for. `#273` and `#317` are in the base now, and the phantom deletion is gone: `git diff --stat main HEAD` and `git diff --stat main...HEAD` agree at **13 files, +543, 0 deletions**.

No content change in this round, and that is checkable rather than asserted — `git range-diff` against the previous head shows both commits `=`, re-parented only:

```
1:  2ec51ce = 1:  5589dfb feat(output-guard): last-line invariant on the response path — restore, log, never corrupt
2:  3c4ecfa = 2:  aa882b0 output-guard: the forwarded tail must stay assistant-terminal-equivalent
```

Tests on the new base: **1768 tests / 56 suites / 1767 pass / 0 fail / 1 skipped** (~25 s). The count is up from 1724 last round because the new base carries the tests from `#273`/`#317`. The guard's own 15 tests are green, verified by title rather than by an exit code.

On the CI failures — agreed, and one detail from that same run supports your reading: `test (20)` **succeeded**, all nine steps green including `Run tests`, while `test (18)` and `test (22)` each have exactly one step (`Set up job`) that died at *Failed to resolve action download info — Service Unavailable*. So the suite did execute on the old head under Node 20 and passed; only those two legs never reached it.

https://claude.ai/code/session_01JS8gNw5DG83rqBvo9wzaH5

cnighswonger (approved):

PR #279, opened 2026-07-29
fix(thinking-block-sanitize): protect continuations by shape, not by tail distance

Gunther-Schulz:
`planSanitize` protects a tool-continuation's thinking only while it is the **latest** assistant turn (`i === latestAsst`). But the property that makes it protected — "terminal `tool_use` answered by a following `tool_result`" — belongs to the message, not to its distance from the tail. The moment a later turn lands, the same byte-identical message flips from protected to stripped: a mid-history mutation the proxy itself causes on every request where a continuation ages out of the tail — i.e., exactly the whole-prefix cache re-write this extension exists to prevent.

**Measured** (offline cross-request stability check over live captures, per-extension attribution): 133 violations over 563 requests on one session, 76 over 169 on another — every one attributed to this extension. Zero after the fix.

The fix: drop the `latestAssistantIndex` gate; protect any `isActiveToolContinuation(messages, i)`. A continuation deep in history keeps its thinking exactly as first sent — byte-stable, and the shape the API already accepted. Stripping it later buys nothing (the 400 this extension prevents concerns the latest turn) and costs a full prefix re-write.

**Regression test** pins the `v2StripSigned` arm — the v1 path only drops omitted thinking, so signed thinking is where the old gate bit. Against unpatched `main` the new test fails; with the fix, all 47 pass.

Standalone — no dependency on the #272–#278 series (the measurement tooling that found it is #276).

### DIFF
--- a/proxy/extensions/thinking-block-sanitize.mjs
+++ b/proxy/extensions/thinking-block-sanitize.mjs
@@ -100,9 +100,11 @@ function answersToolUse(msg, toolUseId) {
 // narrow as the approved rule: an unanswered terminal tool_use, or a later
 // tool_result that answers a *different* call, is not the protected case.
 //
-// The predicate is a function of the message and what follows it, NOT of its
-// distance from the tail — see planSanitize's stability note for why that
-// distinction is load-bearing.
+// The predicate ITSELF is a function of the message and what follows it, NOT
+// of its distance from the tail — see planSanitize's stability note for why
+// that distinction is load-bearing for v1. (planSanitize additionally scopes
+// v2's use of this predicate to the latest turn only — that's a decision
+// planSanitize makes on top of this predicate, not a change to the predicate.)
 export function isActiveToolContinuation(messages, idx) {
   const msg = messages[idx];
   if (!msg || !Array.isArray(msg.content) || msg.content.length === 0) return false;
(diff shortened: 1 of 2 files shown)


Gunther-Schulz:
Pushed `0b67dbf` — the protection is scoped to v1, and v2's directive contract is restored.

`planSanitize` now computes the index of the latest assistant turn and splits the guard:

- **v1** keeps the by-shape protection at any position. Its justification is empirical and unchanged: nothing forces us to touch a byte-identical replayed block once it ages out of the tail, and doing so buys nothing while costing a full re-write.
- **v2** gets the protection **tail-scoped** — only the latest assistant turn, and only when that turn is itself an active continuation. Any earlier active continuation loses its signed/redacted thinking under `v2StripSigned`, exactly as main did before the PR.

The reasoning for the asymmetry is now in the file rather than implied: v2 fires only on a tools-hash mismatch, which has already busted the prefix from that point on, so protecting a deep-history continuation there buys zero cache benefit while reintroducing the stale-signature risk v2 exists to remove (v2 directive, lines 58-67 and 153-159).

Your repro, run literally, and its control:

```
planSanitize([answered continuation, tool_result, later assistant], {v2StripSigned:true})
  → droppedV2: 1, prior turn content ["tool_use"]        // signed thinking stripped, matches main

planSanitize([answered continuation, tool_result], {v2StripSigned:true})
  → droppedV2: 0, prior turn content ["thinking","tool_use"]   // latest continuation still protected
```

**One thing we changed that you did not ask for, flagged because it is a judgment call rather than a mechanical fix.** The PR's own "position independence" regression test used `v2StripSigned: true` to assert that protection survives aging out of the tail — which is precisely the contract your review says is wrong. Left alone, it would have pinned the bug: a green suite asserting the overreach. We reworked it to test v1, which is its actual claim, rather than deleting it — so the historical 133/563 and 76/169 measurements stay attached to a test that is now correct — and added the two v2-contract tests separately.

That generalises, and it is the lesson we took from this one: when a review says a fix is too broad, the PR's own new test may need narrowing too, not just the implementation. A test written to pin a fix can encode the fix's own overreach and then pass, which is worse than no test.

Verification, real output:

- Red-first, arrangement stated: the implementation alone was stashed back to the PR head with the new expectations kept in place (not a working-state revert, which would have removed both and gone vacuously green). The new v2 test failed — signed thinking not stripped — then passed with the fix restored.
- Targeted suite 49/49; full `npm test` **1433/1433**. The first full run showed one failure in `test/proxy-wrapper.test.mjs` (`--remote-control` reads a lowercase-only `no_proxy`), which is this machine's ambient `NO_PROXY`/`HTTPS_PROXY` leaking in — unsurprising, since the machine routes through the proxy this repo builds. Re-run with those unset: clean.
- Hygiene greps over this round's own commit: no matches in added lines or in the commit message.

cnighswonger (approved):

PR #280, opened 2026-07-29
feat(prefix-diff): full cache-key attribution — marker windows, tail snapshots, per-tenant baselines

Gunther-Schulz:
Rework of prefix-diff from a head-only preview into the attribution instrument the rest of the series cites. Read-only throughout — it observes and logs, never mutates a request — so the blast radius is misreporting, not corruption.

## What changed, each step measured

- **Marker-anchored + tail snapshot windows** — long-session busts were invisible to the head-only diff: on every deep request the divergence sat past the preview window, and the log said "no differences" while the bill said otherwise.
- **Full cache-key coverage** — params (`model`, `max_tokens`, `thinking`, `output_config`, `speed`), betas, and the `anthropic-beta` header. Each was a blind spot that produced a false "clean" on a real bust.
- **Per-tenant baselines** — keyed by the agent-id header when present, else the **full** system-prompt hash. The prior 400-char truncation merged agents sharing a long preamble (the shape every real agent system prompt has), silently combining baselines and attributing diffs to the wrong conversation. Co-tenant diffs are labelled as such, never guessed. The truncation bite is pinned: two prompts sharing a ~960-char preamble must not collide.
- **Change-kind recording** — *how* a mid-history message changed (which block, what kind), not only where; the where alone cannot separate a reminder rewrite from a genuine edit.
- **Robustness** — per-key write serialization, timestamp-based eviction, preview of every content shape.

## Evidence

84 tests green. This instrument produced the attribution lines quoted in #272/#273's evidence and in anthropics/claude-code#76606/#81967 — e.g. classifying `system[…]`, `tools[Agent:added,…]`, `messages@0(user)`, `header:anthropic-beta[+…]` per request pair in production. Standalone: no dependency on the other PRs in the series.

### DIFF
--- a/proxy/extensions/prefix-diff.mjs
+++ b/proxy/extensions/prefix-diff.mjs
@@ -11,6 +11,27 @@
 // No request mutation. Fail-open: any I/O error is swallowed. Set
 // CACHE_FIX_DEBUG=1 to also log swallowed errors.
 //
+// Permissions + retention (2026-08-05, reviewer-requested — the rewrite
+// below persists more than the old diagnostic did, and the persistence
+// story had to catch up):
+//
+//   - All three artifacts, and their rotated forms, are written owner-only
+//     (0600) via write-owner-only.mjs, applied at create time so there is
+//     no window where a file is briefly group- or world-readable.
+//   - CONTENT MINIMIZATION. By default this module persists only hashes,
+//     lengths, and indices — enough to say WHICH block/message/index
+//     changed, never the bytes that changed. Full text (system-block
+//     windows, message previews, event-record previews) is stored only
+//     under CACHE_FIX_PREFIXDIFF_CONTENT=1. Enabling it persists
+//     prompt-derived text to disk; it is an explicit, separate opt-in from
+//     CACHE_FIX_PREFIXDIFF itself.
+//   - CROSS-KEY RETENTION. A sweep runs once per process (the first live
+//     request — there is no dedicated boot hook) deleting artifacts older
+//     than 14 days and, if more than 200 session keys remain, pruning the
+//     oldest keys beyond that cap. This bounds the snapshot directory
+//     across sessions; the pre-existing per-key tenant eviction only
+//     bounded the map inside one key's own file.
+//
 // ---------------------------------------------------------------------
 // Design notes — the four blind spots this version closes (2026-07-27)
 //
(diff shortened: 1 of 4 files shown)


Gunther-Schulz:
Pushed wave 2 for this slice — one commit, 69497a3: event records carry their tap-point view marker, single-sourced from the module's own ordering, so a prefix-diff attribution can say which extension's view it measured. Tests: 85 pass / 0 fail. One deliberate surplus vs. the fork: this branch keeps its tenantId full-text-hash collision test, which the fork solved differently — nothing from the fork commit is missing.

Gunther-Schulz:
Pushed `2f96c88` — permissions, content minimization, and the cross-key retention you said was missing.

**Permissions.** Both write sites (`atomicWriteJson`'s tmp write, `appendEvent`'s append) go through the repo's existing write-owner-only helper rather than a second implementation. One gap we found while wiring it: rotation (`rename` to `.1`) preserves whatever mode the pre-rotation file had, and the helper's repair-once-per-process would not reach it if the first append in a process is also the one crossing the rotation threshold. So the sweep repairs every surviving artifact as it passes.

**Retention — your exact gap was "no cross-key GC, TTL, cap, or sweep".** `sweepSnapshotDir` does an age pass (anything older than 14 days, regardless of key count) and then a key-cap pass (oldest keys beyond 200, ranked by newest surviving file). One honest naming point: there is no boot hook to hang it on — `loadExtensions` imports the module and never calls a lifecycle method — so "on boot" is implemented as "the first live request in this process", self-guarded to run once.

**Content minimization, and it went wider than the three categories the design named.** Default `CACHE_FIX_PREFIXDIFF=1` now persists hashes, lengths and indices only. Beyond system-block text, message previews and event previews, we also had to structure the head-5/tail-3 message windows: they stored real message bytes verbatim, and the old truncation capped only `.text` blocks — `tool_use.input` and `tool_result.content` went to disk uncapped, on every changed request. Leaving them would have made "prompt text does not rest on disk" false by construction.

Not gated: `buildToolsSnapshot` and `buildParamsSnapshot`. Tool names, `model`, `temperature` are not prompt text and you did not cite them. Say if that reading is wrong.

**Verification.** The claim here is a negative — "no prompt-derived bytes on disk" — and a passing assertion about which fields are set does not establish it, so it was measured instead. A sentinel string was planted in four places at once (system text, a string-content message, `tool_use.input`, `tool_result.content`), the extension driven for real, and every written file grepped:

```
default mode   -diff.json 0600  hits=0
               -events.jsonl 0600  hits=0
               -last.json 0600  hits=0        -> 0 hits total

CONTENT=1      -diff.json 0600  hits=6
               -last.json 0600  hits=19       -> 25 hits total
```

The second run is the point: it proves the grep can see the sentinel, so the zero in the first run means something rather than being an instrument that never fires. Both runs were re-done independently of the test suite before this push.

Also: mode-bit assertions on each artifact kind against a real tmpdir (no fs mock); a rotation test seeding a 0644 pre-existing events file and asserting the sweep repairs it; sweep tests for the age pass, the 200-key cap (205 seeded, exactly 200 survive), a missing directory, and a control that non-artifact files are left alone. `npm test` **1500/1500**. The new test file was run against the branch base and fails to even import there, confirming it is new coverage rather than a restatement.

**For the record, since the flag persists prompt text:** `CACHE_FIX_PREFIXDIFF=1` alone now persists only hashes, lengths and indices — never prompt-derived bytes. Setting `CACHE_FIX_PREFIXDIFF_CONTENT=1` in addition opts into the richer diagnostic mode and **will write prompt-derived text to disk** — system-prompt block text, message content previews, event-ledger previews — under `~/.claude/cache-fix-snapshots/`, at 0600, subject to the 14-day / 200-key sweep. Enable it only while actively debugging a cache bust.

cnighswonger (approved):

Issue #10, opened 2026-04-12
Interesting read about 1 hour -> 5 minute cache


cowwoc:
You might find these links interesting:
* https://github.com/anthropics/claude-code/issues/46829#issuecomment-4231266649
* https://news.ycombinator.com/item?id=47739260

It's not clear to me how they figure users are actually saving money here, but it seems relevant to your project.

cnighswonger:
@cowwoc — thanks for flagging these. Both are directly relevant.

The Jarred-Sumner response on #46829 is the first official confirmation of the per-request TTL selection mechanism we've been documenting — our blog post ([The 5-Minute Baseline](https://veritassuperaitsolutions.com/5-minute-baseline-tools-array/)) arrived at the same finding from the outside via the `ScheduleWakeup` tool description and cross-version prefix measurements. His economic argument (5m writes are cheaper than 1h writes for one-shot calls) is correct for automated workloads, but inverts for interactive multi-turn sessions where the same prefix is re-read 10-100 times — which is the exact population reporting quota drain.

The HN thread is useful context for where the community temperature is. We're tracking both and will be responding to Boris's post on #45756 and Jarred-Sumner's framing on Monday with data from our interceptor telemetry.

Appreciate you connecting the dots for us.

cowwoc:
For what it's worth, I think the TTL should be configurable for the simple reason that 1 hour might be too long but 5 minutes is way too short. Specifically, I think it takes most people more than 5 minutes to notice that Claude completed a task, review the changes and ask it to merge. This is a subjective decision depending on one's working style and the number of concurrent claude instances you have running. If I had a choice, I would use a cache TTL of 30 minutes <shrug>

Thanks for the hard work. I'm going to go ahead and close this issue.

cowwoc:
I just ran across https://news.ycombinator.com/item?id=47740756

Does this match your understanding? Or are subscription accounts impacted as well?

cnighswonger:
@cowwoc — to your two points:

**On configurable TTL:** agreed, and Jarred Sumner's response on #46829 actually confirms this is by design — the client selects per request. Main conversation turns get 1h, subagent calls get 5m. The issue is that users have no visibility into which tier any given call got, and no way to override it. A user-configurable TTL (or at minimum, a floor) would let people choose their own tradeoff between cache-write cost and cache-hit frequency. Our interceptor does this bluntly by forcing 1h on everything, but a first-party configurable option would be better.

[...]

PR #281, opened 2026-07-29
remove(messages-cache-breakpoint): superseded by canonical content identity (#272 has landed — your call now)

Gunther-Schulz:
> **DRAFT — blocked on #272**, deliberately opened now so the plan is visible during that review (this branch stacks on #272; the replacement is present in it). Will mark ready when #272 merges. Whether to take this at all is genuinely your call — see the trade-off below.

## Why remove it

- Its #47098 rationale (protect the head against mutation) is covered by #272's canonical content identity, which recognises a message across re-serializations instead of pinning index 0.
- It spends the **4th and last** API breakpoint slot (Claude Code occupies three). Holding it isn't free even while the extension behaves: any future consumer of that slot silently starves. On our fork that starvation *masked* a defective mid-history rung scheme for a day — freeing the slot armed it, measured at 57 self-inflicted stability violations before the scheme itself was retired. The ordering rule this removal carries: retire anything breakpoint-dependent **before** freeing the slot.
- Measured after removal on two full-session corpora through the real pipeline: 0 stability / 0 safety / 0 sequence violations, 4th slot deliberately empty.

## The honest trade-off

`messages-cache-breakpoint` ships **enabled**; #272's replacement is **gated off by default**. Merging this as-is changes default behavior for users who don't enable the new extensions. Options, in our order of preference: (a) flip #272's gates to default-on at the same time, (b) take this removal only with a release note, (c) keep the extension but default it off. We run (a)-equivalent in production; the corpora above measured exactly that configuration.


Gunther-Schulz:
Rebased onto #272's new tip (c713d0e, the duplicate-suppression addition) — no changes to this PR's own commit; still stacks on #272, review alongside it.

Gunther-Schulz:
Rebased onto the current #272 tip (force-with-lease; the retire commit rides on top unchanged as b07c0dc). Tests at the new base: 67 pass / 1 designed skip / 0 fail. Merge order dependency unchanged: this lands after #272.

Gunther-Schulz:
Rebased onto current `main` (`4ab9cf8`) and taken out of draft. This PR's dependency landed ten days ago and I hadn't noticed — that's on me.

**Why it was conflicting.** #272 merged on 2026-08-05 as a squash (`b00b141`), so its eight original commits are not ancestors of `main`. This branch was still replaying seven of them, which is what `mergeable: CONFLICTING` was reporting — not a real content conflict. The branch is now the single removal commit cherry-picked onto `4ab9cf8`, plus one new docs commit. Diff is 10 files, +6/−1089.

**Safety check before freeing the 4th `cache_control` slot.** The removal's own rationale carries a precondition — retire any breakpoint-dependent scheme *before* freeing the slot — so I checked it against `main` rather than against the fork it was written on. Two independent reasons it holds here:

1. Injection is opt-in and default-off (`CACHE_FIX_INJECT_MESSAGES_BREAKPOINT` is unset in everything shipped). The 4th slot is therefore already free in default operation, so removing the extension cannot newly arm anything that wasn't already armed.
2. The scheme the precondition warns about — the mid-history breakpoint ladder — is already gone from `main`; `insertion-normalization.mjs`'s header records it as removed.

Behavioural impact: operators who never opted in see no change. Operators who did set `=1` lose breakpoint-#3 injection and regain the free slot.

**Second commit: the docs cascade, which the original removal missed.** The removal deleted code and tests but left the extension documented as live in every user-facing surface — all three READMEs carried a "Cache breakpoints" section with working `export` lines, and both `docs/monitoring.md` and `docs/extension-impact-guide.md` listed the two env vars as current. Anyone following those would have set a variable nothing reads. Search used, after the removal: `grep -rlnE 'messages-cache-breakpoint|CACHE_FIX_INJECT_MESSAGES_BREAKPOINT|CACHE_FIX_DUMP_MESSAGES_HEAD'` over the tree — 15 files before, 10 after, and the 10 remaining are all dated records (CHANGELOG history, six `docs/code-reviews/`, two `docs/directives/`).

The impact guide's sections are ordinally numbered, so dropping #10 renumbered the two below it (`microcompact-stability` 11→10, `read-dedupe` 12→11). CHANGELOG gets a `Removed` entry under `[Unreleased]`.

**One judgment call for you.** I left `docs/directives/` alone — both the extension's own spec and the passing mention in `proxy-ttl-tier-detection.md`'s interaction table — treating directives as dated design records rather than live reference, the same status as `docs/code-reviews/`. If you'd rather they carry a superseded marker, say so and I'll add it.

**Verification.** Full suite green at the pushed commit: 1741 tests, 1740 pass, 0 fail, 1 skipped (the pre-push hook runs it at the pushed SHA, so that's the tested commit, not an earlier one). The earlier CI failure on this branch — `28. read-dedupe loads at order 380…` — needed no fix: it failed against the old base where the assertion spanned the removed extension's order, and `main` has since reworded it to reference only 370 and 400. Confirmed by running that file on `4ab9cf8` before the cherry-pick (42/42 green) and again after (42/42 green).

— Claude Opus 5, agent for @Gunther-Schulz

Gunther-Schulz:
Rebased onto `main` — conflict-free and mergeable again. The only conflict was
the `Unreleased` CHANGELOG section, where your `### Changed` entry and this
branch's `### Removed` entry both landed; both are kept.

**Your deferral condition is met.** You wrote: land #272, run with it, then
decide with evidence rather than in advance. #272 merged 2026-08-05 and has been
running since.

One addition beyond the rebase: `README.fr.md` (#334) landed after this branch
was last rebased and still documented the two removed env vars, while
`README.md`/`zh`/`ko` no longer did — a race between two same-day PRs, not a
missed file. The retirement now cascades to it. The design record at
`docs/directives/proxy-messages-cache-breakpoint.md` gains a superseded callout
rather than being deleted: an unmarked retired spec reads as current until the
reader notices `main` does not match it. The passing mention in
`proxy-ttl-tier-detection.md` is deliberately left alone — it only names the
extension.

Still genuinely your call whether to take this at all. If you want the post-#272
evidence in a form you can check rather than take from me, name the form and I
will produce it.

Title fixed — it said DRAFT and the PR was not one.

PR #282, opened 2026-07-30
upstream-change-detection: a conversation growing is not an upstream change

Gunther-Schulz:
## What

`upstream-change-detection` treats a `messages.count`-only diff as routine conversation growth: it updates the baseline silently instead of alarming. Any other changed path still alarms, with the count riding along in that diff.

## Why

Measured on live traffic: **97% of all recorded alarm events were `messages.count`-only diffs** (4,661 events on the day the telemetry verdict first read the log) — routine turns drowning the file this alarm exists to keep meaningful. After the change, an alarm event again means an actual upstream-shape change.

- **Size/complexity budget**: 11 LOC extension change + 59 LOC tests, one file each. No new abstractions.
- **Threat model**: n/a — no new inputs or trust boundaries; the extension reads the same request shapes as before.
- **Maintainability**: pure narrowing of an existing alarm predicate; no dead code, no shims.
- **Performance/reliability**: one extra path comparison per event; negligible.
- **Load-bearing?** Yes — extension code on the wire path, but zero novel logic: the change is which diffs update silently vs. alarm, already reviewed, tested (28/28 including a red run against the old behavior), and serving on the fork.

## Testing

`node --test test/proxy-upstream-change-detection.test.mjs` → 28 tests, 28 pass, 0 fail at this exact commit on top of current upstream `main` (clean cherry-pick, fast-forward-able).

### DIFF
--- a/proxy/extensions/upstream-change-detection.mjs
+++ b/proxy/extensions/upstream-change-detection.mjs
@@ -463,11 +463,17 @@ async function _processRequest(body, headers, { dir, map, fs }) {
   // messages.count grows on every ordinary turn — a conversation living is
   // not an upstream structural change, and logging it drowned the alarm
   // file (97% of all recorded events were count-only diffs, measured
-  // 2026-07-30 over 4661 events). A count-only diff updates the stored
+  // 2026-07-30 over 4661 events). A count-only INCREASE updates the stored
   // fingerprint silently; any OTHER changed path still alarms, with the
-  // count change riding along in its diff if both moved.
-  const alarmDiff = diff.filter((d) => d.path !== "messages.count");
-  if (alarmDiff.length === 0) {
+  // count change riding along in its diff if both moved. A count-only
+  // DECREASE (compaction, truncation, an upstream rewrite that drops
+  // messages) still alarms — this detector exists to catch shape changes
+  // we didn't anticipate, and a shrink is exactly that kind of event.
+  const isCountOnlyGrowth =
+    diff.length === 1 &&
+    diff[0].path === "messages.count" &&
+    diff[0].to > diff[0].from;
+  if (isCountOnlyGrowth) {
     return { event: "noop", nsKey };
   }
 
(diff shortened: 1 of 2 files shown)


Gunther-Schulz:
Pushed `9474a39` — the suppression is now increase-only, and a count-only decrease alarms again.

`upstream-change-detection.mjs:463-475` previously filtered `messages.count` out of the diff and treated an empty remainder as a noop, which swallowed the decrease along with the increase. The predicate is now explicit:

```js
const isCountOnlyGrowth =
  diff.length === 1 &&
  diff[0].path === "messages.count" &&
  diff[0].to > diff[0].from;
```

A count-only shrink — compaction, truncation, an upstream rewrite that drops messages — is exactly the kind of unanticipated shape change this detector exists to catch, so it alarms. Any multi-path diff alarms as before.

Verification, real output, and the red-first arrangement stated because it is the part that is easy to fake:

1. **Red first.** The new test 18 was written against the branch head's *unmodified* implementation and run before the fix existed — not by reverting the working state, which would have removed the test along with the fix and produced a vacuous green. `node --test test/proxy-upstream-change-detection.test.mjs` → 29 tests, 28 pass, **1 fail**: `AssertionError: a count-only decrease must still alarm`, actual `undefined`, expected `true`. The old predicate swallowed it, which is the defect.
2. Same command after the fix → 29/29.
3. `npm test` → 1433/1433. The first run showed one failure in `test/proxy-wrapper.test.mjs` ("`--remote-control` reads a lowercase-only `no_proxy`") which turned out to be this machine's own ambient `NO_PROXY`/`HTTPS_PROXY` leaking into the test — unsurprising here, since the machine routes its traffic through the very proxy this repo builds. Re-run with `env -u NO_PROXY -u no_proxy -u HTTPS_PROXY -u https_proxy` → 1433/1433 green. Unrelated file, untouched by this change.
4. Hygiene gate on the diff against `upstream/main` → no matches.

One thing deliberately *not* added: `typeof` guards around `from`/`to`. `messages.count` is always `arr.length` at line 313, so a numeric guard there would be defensive handling for a case that cannot occur.

cnighswonger (approved):

PR #303, opened 2026-08-04
fix(proxy): survive a reload instead of cutting the response

codeslake:
## The failure

A reload of the proxy cuts responses that are still streaming. Three live
sessions on a shared host died this way in one window, each surfacing as
`Connection closed mid-response`.

It is structural rather than a race. The only reload available is
kill-then-respawn, because a successor cannot bind the port until the old
process is gone — so every body still streaming through the old one dies with
it. The 5 s force-close makes it certain: a `/v1/messages` stream outlives that
window by design.

## Measured

Two real server processes, a 12-chunk stream through the proxy, reloaded
mid-flight:

| | chunks | ended | error |
| :-- | --: | :-- | :-- |
| before | 18 | `false` | **`ECONNRESET`** |
| after | 24 | `true` | none |

## The change

**`SO_REUSEPORT` on the listen.** The successor is already accepting before the
predecessor stops, so a reload becomes a handover: the port is never unbound,
and the old process serves what it already accepted.

`listen({port, host, reusePort})` rather than `listen(port, host)` — the option
form is the only one that carries the flag.

**Shutdown grace 5 s → 120 s**, overridable via `CACHE_FIX_SHUTDOWN_GRACE_MS`.
5 s was the right number when a longer wait meant a longer outage — the port
stayed unbound until the process died. With a successor already serving, waiting
costs only this process's own lifetime, and 5 s is far shorter than the one
request a reload must not cut.

## Platform behaviour, measured on three machines

`SO_REUSEPORT` semantics are not uniform, and they are the opposite of what this
PR first claimed. Two listeners on one port, 40 connections, every row with a
plain-bind control that is refused (so a green is not a silent no-op):

| host | kernel | predecessor | successor | |
| :-- | :-- | --: | --: | :-- |
| Linux | 6.8.0 | 20 | 20 | load-balanced |
| macOS | Darwin 24.6.0 | 0 | 40 | last binder takes all |
| macOS | Darwin 23.5.0 | 0 | 40 | last binder takes all |

*(macOS rows contributed by a peer session with access to those machines.)*

Linux splitting new connections between the two listeners would be a hazard if
the predecessor were dying while still accepting. It is not: `close()` removes
it from the load-balance set immediately, and `close()` is the first thing the
SIGTERM handler does.

| Linux 6.8.0 | predecessor | successor |
| :-- | --: | --: |
| both listening | 9 | 11 |
| after `close()` | **0** | 20 |

So the split window spans only the period when both are healthy — a connection
landing on the predecessor there lands on a working process, not a dying one —
and it ends at `close()`. That is also why the longer grace is free: the
predecessor is invisible to new connections for the whole 120 s and is only
finishing what it already accepted.

**A successor that binds and then crashes cannot strand the port.** Measured on
all three, with the successor asserted to be serving first, so `B=0` afterwards
is a transition rather than a pre-existing zero:

| | predecessor | successor |
| :-- | --: | --: |
| both listening | 4 | 16 |
| successor gone | **20** | 0 |

The predecessor resumes. This is the property that matters most for a handover,
and it holds on both platforms.

## The test

Two real `spawn`ed server processes, not two `startProxy()` handles. The
question is whether two PROCESSES can hold one port at once, and an in-process
test cannot ask it. It streams through the first, starts the second, `SIGTERM`s
the first, and asserts the body completes.

Gated on a **capability probe** — it binds two listeners for real and reads the
result — rather than a version comparison. `reusePort` landed in node 22.12, so
18 and 20 ignore the option, listen successfully, and refuse the successor
`EADDRINUSE`; those runtimes keep today's kill-then-respawn reload, which is a
real limitation rather than a test failure. A version compare would answer a
different question: it would skip on a new runtime over an old kernel and run on
an old runtime over a new one.

Three ways, all measured:

| | result |
| :-- | :-- |
| capable runtime | runs, passes |
| probe forced incapable | **skips**, does not fail |
| `reusePort` removed from the server | **fails** |

Full suite `1500/1500`, 0 skipped, on node v24.11.1.

## Not claimed

One `/health` probe at the instant of `close()` can still be refused —
connections sitting in the old listener's accept queue are reset, which is a
known `SO_REUSEPORT` property. That is a new connection failing and being
retried, not an in-flight body being cut, and the latter is the failure this
change was written to remove. A queue-draining handover would need its own
measurement.

### DIFF
--- a/proxy/server.mjs
+++ b/proxy/server.mjs
@@ -528,6 +528,29 @@ function removeSelfHeal() {
  *   await startProxy({ port: 0 })                    // OS-assigned port
  *   await startProxy({ port: 0, watch: false })      // embedded, no fs.watch
  */
+// Is a DIFFERENT-MODE cache-fix proxy already on this port? `reusePort` lets a
+// successor co-bind, which is the point, but a plain proxy and a forward proxy
+// sharing a port makes the kernel round-robin CONNECT between one that speaks it
+// and one that does not. Asks the incumbent rather than predicting it — the
+// answer is on `/health`, which every version of this proxy has served.
+//
+// Returns FALSE on any doubt: nothing listening, a timeout, a non-proxy service,
+// unparseable JSON. A guard that blocked startup on an unanswered probe would
+// turn a slow box into a proxy that will not start, which is worse than the
+// mismatch it prevents.
+async function modeConflict(port, bind, weAreForward) {
+  const body = await new Promise((res) => {
+    const req = http.get({ host: bind, port, path: "/health", timeout: 1000 }, (r) => {
+      let b = ""; r.on("data", (d) => (b += d)); r.on("end", () => res(b));
+    });
+    req.on("error", () => res(null));
+    req.on("timeout", () => { req.destroy(); res(null); });
+  });
+  if (!body) return false;
+  try { return JSON.parse(body).forward_proxy !== weAreForward; }
+  catch { return false; }
+}
+
 export async function startProxy(options = {}) {
   const port = options.port ?? config.port;
   const bind = options.bind ?? config.bind;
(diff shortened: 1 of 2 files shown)


codeslake:
Closing this in favour of #296, which now carries the same work.

Both changed `proxy/server.mjs`, and keeping them apart meant every review had to
reason about two diffs that touch one file. The reload fixes are now commits on
#296 instead.

Nothing is dropped — all five commits moved across:

- `SO_REUSEPORT` on listen so a successor binds the port while the predecessor is
  still serving. Measured: kill-then-respawn leaves the port dead 1073 ms;
  start-first with the option leaves it dead 0 ms.
- macOS throws `ENOTSUP` on the *first* listen carrying `reusePort` (Darwin
  23.5.0 and 24.6.0, node 25/26). Unguarded, the proxy would not have started at
  all there. Retried once without the option, on `ENOTSUP` only — retrying
  `EADDRINUSE` would silently bind a port another proxy already holds.
- A forced shutdown reset in-flight clients. `closeAllConnections()` destroys the
  socket and the kernel answers RST, so a client that had received every byte
  still surfaced `ECONNRESET` and discarded the data. Ending the open responses
  sends FIN: measured `ECONNRESET` → clean EOF.
- A mode guard so two proxies in *different* modes cannot share a port, since
  `reusePort` removes the `EADDRINUSE` that used to prevent it.

Issue #292, opened 2026-08-01
test/fixtures/cc-transcript-shape-snapshot.json carries capture-derived identifiers (public history)


Gunther-Schulz:
While building a mechanical absence scanner for conversation-derived fixtures (part of the fixture-leak remediation coordinated on PR #272), it flagged a second file, on `main` since `16ad235`:

`test/fixtures/cc-transcript-shape-snapshot.json`

Findings by class — deliberately not quoting values here:

- session / prompt / parent **UUIDs** (capture-session identifiers)
- one **448-char thinking signature**
- one ~130-char **raw content string** at `$.source`

None of these are credentials; the exposure class is correlation/identification of the capture session. Since public git history can't be scrubbed (the same reasoning as the claude-code-meter#19 precedent in this repo's hygiene notes), the practical remediation is forward-looking: replace the fixture with a tokenized/synthesized equivalent and treat the leaked identifiers as burned, rather than attempting history surgery on `main`.

If useful: the scanner is `tools/absence-scan.mjs` on our fork (importable single source for the classes; it also runs as a pre-push guard on our side), and the directive delivered on PR #272 proposes a CLAUDE.md hygiene-class addition covering conversation/capture-derived data alongside the existing origin-server class.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

— Claude (drafted for and approved by Gunther Schulz)

Gunther-Schulz:
You were right on every count, and the `_note` was the worst part of it —
a file that says it has been redacted stops being looked at.

Rebuilt on our fork rather than scrubbed, deliberately: a scrub removes
the hazards someone thought to enumerate and ships the unanticipated
field by default, while a file assembled from known-safe values ships
nothing that was not deliberately placed. What survived from the
original is only the four `*_keys` arrays — the observed key sets, which
are the fixture's actual value and the only thing
`test/proxy-session-mirror-envelope.test.mjs` reads. Both `*_sample`
objects are now authored from scratch: synthetic identifiers, a short
placeholder in place of the thinking signature, a generic `cwd` and
branch, and a fabricated `gh issue view --comments` payload that keeps
the nesting the sample exists to document. The false `_note` is gone;
the new one says plainly that the samples are synthetic and names the
two places where the shape deliberately departs from a real record.

Counts, since they are worse than the issue estimated: the old content
produced **10** findings under the scan — 9 UUID-class identifiers
(including one inside the `$.source` transcript path) and the 448-char
[...]

Issue #284, opened 2026-07-31
Index for the 11 open PRs from the cache-fix fork — merge order, sizes, and what's soaked


Gunther-Schulz:
I opened 11 PRs on 2026-07-29/30 without checking appetite first. That was the wrong order — #248-before-#251 is the pattern that works here, and I skipped it. This issue is the retroactive version: one place to read, so you can say yes, no, or "not this one" at the design level without opening eleven diffs.

**Nothing here is urgent from my side.** It all runs in production on my own machines and will keep doing so whether or not it lands upstream. If a slice isn't wanted, say so and I'll close it — that's a useful answer, not a loss.

## The four small independent ones

These touch no shared abstraction, depend on nothing else in the series, and are mergeable in any order. If you only want to look at part of this, look at these.

| PR | What | Code |
|---|---|---|
| #274 | `preForward` mutates `reqCtx.headers`, but `forwardRequest` reads the original `clientReq.headers` — header mutations never reach the wire. In-tree victim: auto-1m-guard's strip mode (its unit test asserts the ctx mutation; nothing proved it forwarded). | +187 |
| #277 | `server.close()` never resolves under a live session (the streaming response is always in-flight), so the 5s watchdog is the *normal* systemd stop path — every `systemctl stop` logs `status=1/FAILURE` and trips `Restart=on-failure`. Force lingering connections, exit 0. | +100 |
| #279 | `planSanitize` protects a tool-continuation only while it's the latest assistant turn — so a byte-identical message flips from protected to stripped as it ages out of the tail. That's a mid-history mutation the proxy causes on every request, i.e. the exact bust this extension exists to prevent. Measured: 133 violations / 563 requests on one session, 76 / 169 on another; zero after. | +70 |
| #282 | `upstream-change-detection` alarms on `messages.count`-only diffs — routine conversation growth. 97% of recorded alarm events (4,661 in one day) were this. Narrows the predicate so an alarm means an actual shape change. | +70 |

## The main series

Two mitigations plus the machinery that proves them, in dependency order:

- **#272 insertion-normalization** (+2,912 code, +15,335 fixture) — CC re-serializes old `<system-reminder>` blocks mid-history, re-billing everything after the edit index. Same diagnosis independently reported in anthropics/claude-code#76606. Pins volatile blocks to their first-seen serialization, keyed by content identity rather than position.
[...]

PR #306, opened 2026-08-05
absence-scan: standalone fixture-hygiene scanner + push-time CLI

Gunther-Schulz:
Splits `tools/absence-scan.mjs` and its test out standalone, so the scanner can land independently of the PRs that need it. Asked for in #284 and again in #292.

Ref #302
Ref #292

## What it is

An importable scanner plus a CLI for five fixture-hygiene classes — a base64 payload run over 200 chars, a raw nested wire payload, a live wall-clock timestamp, a capture UUID, and untokenized capture prose — plus a filename-shape check for the `s-<8hex>` raw session prefix. `--git-range <old>..<new>` scans only what is newly reachable at a ref, which is the shape a pre-push hook needs.

Findings never echo the match. A leak reporter that prints the leak into a terminal or a CI log has moved it, not found it, so a finding carries the class, the file, the JSON path and a length — never the bytes.

## Known blind spot, stated because it changes how you would wire it up

In `--git-range` mode the candidate list is filtered by `SCANNABLE = /\.jsonl?$/i` **before any class runs**. A capture identifier in a tracked `.mjs` or `.md` is invisible to it, whatever the class definitions say — proved here with a planted UUID that the `.json` path caught and the `.mjs` path did not. If you wire this into a push hook, pair it with a source-file grep until that is addressed. We would rather ship the limit written down than let it be discovered later.

## Two findings against this repository's current tree

Neither is a defect in the scanner, and neither is ours to fix, so they are reported rather than patched:

1. **`test/fixtures/cc-transcript-shape-snapshot.json` still carries real capture content here.** Running the scanner over this tree returns 10 findings, exit 2 — a real session id, a ~450-character thinking signature, a real cwd and branch, and verbatim third-party GitHub comment bodies. That is #292, and our fork's copy has since been rebuilt from known-safe parts; this repository's copy has not. Offer stands from #292: we are happy to send the rebuilt fixture as its own PR, or leave it to you.
2. **Several UUIDs under `docs/code-reviews/` and `docs/directives/`** are flagged by the source-tree guard our fork runs. We could not determine whether they are real or synthetic, and did not guess.

## What the suite covers, and what it deliberately does not

Every class has a seeded defect that must make exactly that class fire and no other — the extraction's own guarantee, since a class asserting the ABSENCE of a defect over a clean corpus passes just as well when it has been neutered.

Two tests our fork keeps here are **deliberately not ported**: one asserting its transcript fixture is clean, one walking the source tree against a synthetic-UUID roster. Both are guards over the host repository's *content* rather than the tool's behaviour, and both fail against this tree for exactly the reasons above. A tool's bite must go red on the tool's defects; a suite that also goes red on its host's data cannot be landed by whoever adopts it, and softening it to pass would be worse. The reason is written where the next reader will look. Adopting repos wanting the source-tree guard should add it with their own roster.

## Boundary conditions you named

- **No `test/fixtures/harvested/` directory**: the corpus scope reads as empty-and-passing rather than erroring, and — pinned by a new test — does not widen into skipping the byte-level classes, which need no corpus to be true.
- **No `cc-transcript-shape-snapshot.json` allowlist entry.** The fork retired it once the fixture stopped needing an exemption; only the `LEDGER-*.json` entry remains.
- **The test's scratch-repo helpers scrub git's environment** (`GIT_DIR`, `GIT_WORK_TREE`, `GIT_INDEX_FILE` and four more, set to undefined rather than `""` — an empty string is still "set" to git).

That last one is not hypothetical, and it is the reason it ships this way rather than as advice about how to invoke the suite. Git exports `GIT_DIR` into hooks — relative for a main-tree push, absolute for a worktree push — and a scratch repo spawned with `cwd: tempdir` but an inherited environment is not scratch at all: every `git init` and `git config` resolves to the invoking repository. Run from a git hook, the unhardened suite writes `core.bare = true` and a fixture identity into the caller's real config and breaks their work tree. We reproduced it, twice, on our own repository — the second time because the first fix hardened the *hook* rather than the spawn, which closed one caller and left the hazard. Measured either way: unhardened, the repro moves an enclosing repo's config md5; hardened, 15/15 pass and the md5 is byte-identical before and after.

### DIFF
--- a/test/absence-scan.test.mjs
+++ b/test/absence-scan.test.mjs
@@ -23,13 +23,20 @@ import { tmpdir } from "node:os";
 import { join, dirname, sep } from "node:path";
 import { fileURLToPath } from "node:url";
 
-import { scanDocument, scanContent, isAllowlisted, CLASSES } from "../tools/absence-scan.mjs";
+import { scanDocument, scanContent, isAllowlisted, CLASSES, NAME_UUID_PREFIX } from "../tools/absence-scan.mjs";
 
 const TOOL = join(dirname(fileURLToPath(import.meta.url)), "..", "tools", "absence-scan.mjs");
 const CORPUS = "test/fixtures/harvested";
 
 // Synthetic, and shaped like the thing each class is defined against.
-const FAKE_UUID = "0123abcd-4567-89ef-0123-456789abcdef";
+// Assembled from parts rather than written as a literal: this file is
+// scanned by the very tool it tests, and a UUID-shaped literal in source
+// is exactly what the widened source scan is built to find. Joining the
+// groups keeps the VALUE identical for every assertion below while the
+// source text carries no identifier shape. The alternative — an exemption
+// naming this constant — was tried and discarded: it is the constant the
+// leak bites plant, so exempting it left three of them green.
+const FAKE_UUID = ["0123abcd", "4567", "89ef", "0123", "456789abcdef"].join("-");
 const LONG_B64 = "QUJDREVGR0hJSktMTU5PUFFSU1RVVldYWVphYmNkZWZnaGlqa2xtbm9wcXJzdHV2d3h5ejAxMjM0".repeat(4);
 const TOKEN_TEXT = "t_0123456789ab_42";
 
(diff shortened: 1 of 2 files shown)


Gunther-Schulz:
Thanks for putting Chris's three calls here rather than leaving them to be inferred. Answering the ask, delivering the UUID list you asked for, and adding one measured data point for Q3. Apologies for the delay — the round sat unread on our side for eight days.

## The ask: we keep the scanner, we do not take the installer

**#306 stays ours and lands as-is** — it is already open and `MERGEABLE`, so a split-out on your side would buy a fork/rescan cycle for the same commits. No credit needed; it is the same work either way.

**The hooks installer + wrapper + `.githooks/pre-push` we would rather NOT own.** Not reluctance about the work — it is that we cannot test it where it matters. The installer has to compose with visits-01's existing `post-merge` and with whatever else lives in `.git/hooks/` on your machines, and none of that exists here. We would be writing shell against a description of your environment and you would be the one finding out where it was wrong. The scanner is a pure function over bytes and travels; hook plumbing does not.

So: you build the installer against whichever scanner version lands, and we answer questions about the scanner's behaviour at any depth you want.

## Q3 — a measured corroboration, from having made exactly that mistake

`core.hooksPath` is single-valued per repo, so it does not chain: setting it *replaces* whatever `.git/hooks/` held. We did that to ourselves on 2026-08-05.

Our machine runs a global pre-push dispatcher (via a global `core.hooksPath`) that carries the fixture-leak scan for every guarded checkout, and then chains the repo's own `.git/hooks/pre-push`. Setting a **repo-local** `core.hooksPath` in the fork checkout replaced that dispatcher outright. The repo's own suite hook kept running, every push kept passing, and the leak scan — the one guard standing in front of unscrubbing public history — was simply gone. Nothing looked broken, which is the whole problem: a guard removed by configuration reports exactly what a guard finding nothing reports.

The repair was chaining, not pointing: a symlink for the repo's own hook inside `.git/hooks/`, plus a check in our machine bootstrap that now **fails** on any repo-local `core.hooksPath` in that checkout. So Chris's Q3 call matches what the failure mode actually costs, from the other direction: the installer preserves what is already installed, `core.hooksPath` silently evicts it.

## The other UUIDs, as asked

Scanned `main` (`4ab9cf8`) with our fork's current scanner, which reaches source files:

```
node tools/absence-scan.mjs --at <ref> $(git ls-tree -r --name-only <ref> -- docs/)
scope: 287 source file(s) — capture-key-prefix and capture-uuid only
16 finding(s)
```

**Four are already closed by #319** — `docs/code-reviews/pr-299-round-1-codex.md:18,23` and `pr-299-round-2-codex.md:12,13` now carry the synthetic replacements. Our class fires on any canonical UUID and cannot tell a synthetic from a real one by construction, so these are expected false positives on a scrubbed file, not a missed scrub.

**Six carry an all-zero first segment** and read as deliberate placeholders — no action indicated:

```
docs/code-reviews/README.md:11
docs/release-tests/pr-220-image-retry-circuit-breaker-parallel-harness-2026-06-11.md:47, 67
docs/release-tests/pr-221-jsonl-session-mirror-parallel-harness-2026-06-12.md:35, 57, 114
```

**Six we cannot classify, and did not guess** — this is the set worth your eyes:

```
docs/benchmarking.md:45
docs/code-reviews/proxy-owned-oauth-refresh-impl-r1-codex.md:54
docs/directives/proxy-cache-warmer-v3.7.0.md:156
docs/directives/proxy-cache-warmer-v3.7.0.md:168      (same value as :156)
docs/directives/proxy-jsonl-session-mirror.md:65
docs/directives/proxy-quota-status-per-session.md:176
```

Five distinct values across five files. Whether any of them is a real transcript id is a question only you can answer; we are reporting positions, never values, for the same reason the scanner does.

Worth naming: **the scanner in this PR cannot see any of the twelve.** In `--git-range` mode it filters candidates to `.jsonl?$` before a class runs, which is the blind spot this PR's body documents — every one of these findings comes from the widened version, which is the next section.

## The two agreed changes are now IN THIS REF — `20365e1`

Both were missing from `fb9763b`, and they were never in the #276 ref either. We reported them there as landed while reading our fork's copy of the file rather than the pushed ref; that is corrected on #276 with the measurement. So this time, read back out of the ref rather than out of a working copy:

```
$ git fetch upstream +refs/pull/306/head:pr-306-head && git rev-parse --short pr-306-head
20365e1
$ git show pr-306-head:tools/absence-scan.mjs | wc -l
479
$ git show pr-306-head:tools/absence-scan.mjs | grep -n 'NAME_UUID_PREFIX =\|SOURCE_SCANNABLE'
113:export const NAME_UUID_PREFIX = /(^|[^0-9a-zA-Z])s-[0-9a-f]{8}(?![0-9a-f])/;
291:  if (SOURCE_SCANNABLE.test(file) && !SCANNABLE.test(file)) {
343:const SOURCE_SCANNABLE =
```

Red-first, both arms, so the bites are known to discriminate rather than merely pass: with `fb9763b`'s scanner restored under the new tests, **16 pass / 3 fail** — the boundary bite, the source-scan bite, and a git-range bite that plants an identifier in a `.mjs` and in an extensionless hook script. Against the change, 19/19. The three that fail are the three that must; the sixteen that pass are the baseline, which is what makes it a split rather than a module-load error.

### Two things the widening turned up that are worth more than the diff

**It fires on the suite's own synthetic identifiers.** The obvious repair is a declared exemption naming that constant — and here that is the *wrong* repair, which we found by making it. The value a source scan trips over is the very value the leak bites plant, so the exemption silently turned three of the tool's own red-first cases green. An exemption that blinds an instrument to its own defect class is worse than the false fire it removes. The suite now assembles its identifier-shaped constants at run time instead: the source text carries no such shape, nothing is blessed by name, no predicate is softened, and the scanner is clean on its own repository.

**The same thing bit our push gate, which is the general lesson.** Our own pre-push runs this scanner, and the review thread's test vectors — the word-tail negatives and the real-shape positives alike — are themselves identifier shapes. Those are assembled too, and the comment above `NAME_UUID_PREFIX` now *describes* its examples instead of writing them out. A guard against identifier shapes in source means test vectors and documentation get constructed rather than quoted. Worth knowing before wiring this into CI, because it will apply to your fixtures the same way.

**One consequence to name rather than ship quietly:** `SOURCE_SCANNABLE` partly *closes* the blind spot this PR's body documents, which moves the boundary you drew between #306 and #302. The wrapper's source-file grep becomes a belt over the scanner rather than the only thing covering that class. It also means that once this lands, a `--git-range` run touching `docs/` sees the twelve findings listed above rather than none of them. Both are improvements, but they change what #302 is compensating for, so they belong to you as a decision rather than arriving as a side effect — say if you would rather the widening land as a follow-up PR after the installer, and we will lift it back out.

### One unrelated observation, since we ran your suite

`test/install-service.test.mjs` binds a hardcoded `127.0.0.1:9876`, so it fails for anyone whose machine already holds that port — here a local QGIS MCP server did, and it blocked our own push until the port was freed. Our fork takes a free port from the OS instead (`createServer().listen(0)`, then assert on the number passed in; `--proxy-port 0` is not usable, because the assertion needs the value). Happy to send that as its own one-file PR — flagging rather than riding it in here, since you asked to keep unrelated changes out of PRs.

cnighswonger (approved):

PR #304, opened 2026-08-04
fix(proxy): survive a reload instead of cutting the response

codeslake:
## What

A reload of the proxy cuts every session on the port. This makes the listening socket outlive the process that serves it, and makes the incumbent's forced shutdown end responses cleanly instead of destroying them.

## Why it is user-visible

A reload today is kill-then-respawn, so the port is unbound for the gap. Measured: a stream through this proxy died `ECONNRESET` after 18 chunks; with the socket held across the handover the same stream ran to completion.

The forced-shutdown path had a second cut. `closeAllConnections()` destroys the socket and the kernel answers RST — measured, a client that had already received every byte still surfaced `ECONNRESET` and threw the delivered data away. `res.end()` sends FIN, which the same client reads as a clean EOF.

## What is in it

- **A supervisor holds the listening socket and hands it down on fd 3** (`LISTEN_FDS`). The proxy serves a socket it neither binds nor closes, so a reload replaces the serving process while the port stays bound. When fd 3 is not servable — in an IPC-forked child it is the IPC channel and `listen` fails `EEXIST` — it falls back to binding its own port, because a degraded proxy beats no proxy.

  This **replaces an earlier `SO_REUSEPORT` co-bind**, which worked on one of the three runtimes we run:

  | runtime | `reusePort` |
  | --- | --- |
  | node ≥ 22.12, Linux | honoured; successor co-binds |
  | node 18.20.8 / 20.20.2 | ignored; successor gets `EADDRINUSE` |
  | node 25/26, macOS | throws `ENOTSUP` on the **first** listen |

  An inherited socket needs no platform support and has exactly one listener, which also retired the mode-conflict guard the co-bind design required.

- **`bin/gap-relay.mjs`** (new, 293 lines): a standby armed by the holder. It is spawned holding a descriptor to the same listening socket and carries connections to the next hop the instant its holder is gone — `process.ppid !== bornOf`, no probe and no decision to wait for. Measured across holder-and-proxy-both-killed: two 2 s proving windows cost 3,899 ms, three 250 ms windows cost 694 ms, arming immediately costs 3 ms. It does not stand down on its own; yielding is the claimant's decision, made with SIGHUP.

- **FIN, not RST**, on the watchdog path. Open responses are tracked so the forced shutdown can `end()` them, then force whatever did not take the FIN.

[...]

Issue #223, opened 2026-06-13
Directive: statusline served-model divergence indicator (real-time classifier-swap visibility)


cnighswonger:
# Directive: statusline served-model divergence indicator

**Load-bearing?** Yes. This adds the first real-time surface for server-driven model swaps, which today are invisible to operators until they happen to inspect a transcript.

## Why

Today, when CC's safety classifier swaps the served model mid-session (Fable → Opus 4.8, etc.), there is no operator-visible signal that the swap occurred. The user has to either notice the in-TUI banner in real time, or inspect the session JSONL after the fact, or read the `model` field of a usage.jsonl row. None of those is a live surface. The result is sessions like gowy222's six occurrences on CC#66728, where the downgrade is detected hours into a downgraded run instead of when it lands.

The proxy already sees `requested_model` (request body) and served `model` (response). Two fields, one comparison per turn. The statusline already reads from `~/.claude/quota-status/sessions/<id>.json` and renders whatever cache-telemetry writes there. The total work is two small additions (writer + reader) on already-load-bearing components.

This is the first real-time surface for a class of bugs (classifier-triggered downgrades) that has been a recurring upstream issue without a UX path to observe it. Falk-flagged adjacent: this also pairs with the May 9 item [5] (latched-headers observability) and the in-flight extended-cache-ttl-header local test in the sense that both expand the proxy's observability surface to cover server-driven session state CC doesn't expose.

## Functional requirements

1. **Show only on divergence.** Normal-operation default: no extra field in the statusline. When `requested_model ≠ served_model` on the most recent turn, render a divergence indicator.

2. **Sticky state on session-permanent downgrade.** A single divergent turn followed by a return-to-requested could be a transient. A divergence that persists across N consecutive turns or matches the "session locked downgrade" shape from CC#66728 (no recovery without `/model` restore) should latch sticky for the remainder of the session — the bar keeps warning even if subsequent turns happen to align by coincidence.
   - **N threshold suggestion**: 3 consecutive divergent turns or a span of ≥5 minutes, whichever lands first. Reviewers should propose better heuristics if the data suggests them.
   - Sticky state should be observable in the session JSON itself (a `served_model_divergence_sticky: true` field or similar) so other consumers can read it without re-deriving from the request stream.

3. **Short labels with [1m] indicator.**
   - `Fable` / `Mythos` / `Opus 4.7` / `Opus 4.8` / `Sonnet 4.6` / `Haiku 4.5`
[...]

PR #327, opened 2026-08-10
feat(proxy): support per-request upstream base URL override via meta.upstreamOverride

rrva:
### What Changed & Why

Adds optional per-request upstream base URL overriding to the proxy forwarding pipeline via `ctx.meta.upstreamOverride`.

Extensions running in `onRequest` hooks can now target alternative upstream bases per-request (e.g., routing specific model requests to alternative Anthropic-compatible endpoints or custom proxy mirrors) without altering the global `config.upstream` setting.

- `proxy/upstream.mjs`: `forwardRequest()` accepts an optional 4th parameter `upstreamBase` which overrides `config.upstream`. `buildUpstreamUrl` concatenation applies naturally so subpaths (e.g., `https://example.com/anthropic` + `/v1/messages`) are preserved.
- `proxy/server.mjs`: `handleMessages` passes `meta.upstreamOverride` into `forwardRequest()`.
- `test/proxy-upstream.test.mjs`: Added unit tests verifying override behavior and default fallback.

### Example Extension Usage

```javascript
// proxy/extensions/regional-mirror-router.mjs
export default {
  name: "regional-mirror-router",
  description: "Route specific requests to a regional proxy mirror",
  order: 95,

  async onRequest(ctx) {
    const mirrorUrl = process.env.REGIONAL_MIRROR_URL;
    if (!mirrorUrl) return;

    if (ctx.headers["x-use-regional-mirror"] === "true") {
      ctx.meta = ctx.meta || {};
      ctx.meta.upstreamOverride = mirrorUrl; // e.g. "https://eu-west.mirror.example.net/anthropic"
    }
  },
};
```

### Anti-Bloat Metrics
- Production LOC: 6 code lines (+5 comment lines)
- Test LOC: 62 test lines
- Test:Production Ratio: 10.3x

### Non-Functional Checklist
- **Load-bearing?** Yes (modifies core `forwardRequest` egress signature and `handleMessages` parameter handling).
- **Security:** Standard header filtering (`buildUpstreamHeaders`, hop-by-hop stripping) applies unchanged to overridden targets.
- **Back-compat:** `upstreamBase` is optional; when `undefined`, behavior defaults to `config.upstream`.


Gunther-Schulz:
Merge-order note from a neighbouring PR, not a review.

This and #337 both edit `handleMessages` in `proxy/server.mjs` at overlapping
lines — yours at old 169–179, ours at 173 onward — so whichever lands second
will need a rebase. Mentioning it now rather than letting it surface as a
surprise conflict.

No claim on order from my side: #337 is a gated-off optimisation and nothing
depends on it landing first. If yours goes in first I will rebase #337 onto it.

rrva:
Sorry for the silence — I agree with all of it, PB's two questions and AITL's three angles alike. Pushed `1f51c41`: rebased onto current `main` (`78ca948`, so the conflict with `#304` is gone) plus one commit implementing option **(b)**, the proxy-side gate.

## What landed

**`CACHE_FIX_UPSTREAM_OVERRIDE=on`.** I went with the gate rather than (a) because both of you independently arrived at the same trust surface from different sides, and the second side is the one that convinced me: an extension setting this can not only leak the key outbound, it can hand Claude Code a response containing a synthetic `tool_use` block that then executes with my permissions. "Extensions gate themselves" is fine for a knob that changes what gets logged. It is the wrong default for a knob that changes what gets executed. Off means the field is *ignored*, not an error, so an extension that sets it on a host which never opted in is simply inert.

Follows the existing convention: a `config.upstreamOverrideEnabled` getter reading the env per call, and the name added to `PUBLISHABLE_GATES` so `/health` and the capture boot record publish the switch by value.

**Both directions written into the code comment**, not just the outbound half — plus the three opacity items: reverse-proxy-only (forward-proxy's CONNECT path never reaches `handleMessages`), `request-capture` at order 60 recording the pre-override destination, and the `||`-vs-`??` fence left exactly where PB put it.

**One thing the rebase changed that is worth a fresh look**, because it is not in the diff either of you reviewed: `forwardRequest` became `async` in `#304` and grew a hop-resolution preamble. I resolved the conflict so `upstreamBase` feeds `buildUpstreamUrl` *above* that preamble. Consequence, and I think it is the right one: hop resolution, `CACHE_FIX_REQUIRE_HOP` and `buildUpstreamHeaders` all read the *resolved* URL, so an overridden target is subject to the same egress policy as the configured one instead of getting a private path around it. If you would rather the override bypass the chain, that is a deliberate decision someone should make out loud — I did not want to make it silently.

**Tests:** `test/proxy-upstream-override-gate.test.mjs`, four cases at wire level through the real pipeline with a synthetic extension — gate unset, gate set to `"1"` (only the exact string `on` enables it), gate `on`, and gate `on` with no extension setting the field. Mutation-checked: deleting the gate expression fails two of the four. The existing unit tests still cover the mechanism; these cover the decision, which `forwardRequest`-level tests structurally cannot reach.

Full suite locally: 1949 pass, 2 fail — `fixture-reaping` and `proxy-holder-handover`, both on a missing `/proc/net/tcp`, both failing identically on the base commit with my changes stashed. macOS, not me. CI is green on `1f51c41`.

## Question 2, answered straight

The `regional-mirror-router` snippet in the PR description is a sketch of the shape, not code that exists. PB was right to call it aspirational, and I should not have presented it as if a consumer were waiting.

What actually prompted the patch: I was using it locally to route requests for one specific model to a different upstream serving a substitute model. That is a real use, but it is a local unreleased extension — there is no fork branch worth linking, and I am not going to dress it up as a shipping consumer. So the honest state is: **no named consumer**, and the gate is what makes that acceptable rather than something to hand-wave — the seam is inert on every host that has not opted in, including every host that installs an extension which sets the field for its own reasons.

Merge or close is genuinely your call from here. If you would rather carry the change internally per PB's offer, no objection from me and no credit needed.

## One interaction that use turned up

Swapping the served model this way trips the served-model divergence detector (`cache-telemetry.mjs`, issue `#223`): requested ≠ served latches `model_divergence_sticky` immediately across families, or after three consecutive turns within one. For this use it is a true positive — I *did* swap the model — but it presents on the statusline the way an upstream incident presents, and the operator who turned the gate on is the last person who needs telling. Not asking for a change here, and not this PR's job; noting it because anyone who builds the model-routing extension will meet it on day one.

## Merge order

@Gunther-Schulz — thanks for the heads-up. Mine is now rebased on current `main`, so whichever way round is fine; if `#337` goes first I will rebase again.

— rrva's Claude Code session

rrva:
Correction to the line above: "CI is green on `1f51c41`" is only true of GitGuardian and Snyk, which run as apps. The Node 18/20/22 matrix is sitting at `action_required` — a fork PR needs a maintainer to approve the workflow run, and the force-push re-armed that gate. So the matrix has **not** run on this head yet, and my local run (macOS, minus the two `/proc`-dependent files) is not a substitute for it. Whoever picks this up, the run needs approving before the test evidence is real.

— rrva's Claude Code session

PR #337, opened 2026-08-14
feat(proxy): coalesce the duplicate session-start sidecar send (opt-in)

Gunther-Schulz:
Claude Code sends the same session-start sidecar request **twice**, ~6–25 ms apart, byte-identical. Both are answered, and both are charged. This adds an opt-in coalescer that collapses the pair into one upstream call and fans the single response out to both waiting clients.

**Off by default.** It does nothing unless `CACHE_FIX_COALESCE_SIDECAR=1` is set.

## The measurement

Not an inference from the code — this is what one day's traffic on one machine looks like, joined against the completion records by request id:

```
144 duplicate pairs / 114 streaks (longest streak: 11 sends)
258 requests involved
134 of those requests billed
 55 streaks billed more than once
```

The session-start class specifically: haiku, `nMsg=1`, `max_tokens=32000`, arriving at capture lines 3–5, 6–25 ms apart. 47 instances in one measurement window.

The other half of that population is **not** this bug and must not be collapsed: mid-session duplicates are real client retries whose first attempt has no completion record. Suppressing those would leave a genuine request unanswered. That distinction is the whole design constraint here.

## What it does

Four conditions must all hold before anything is coalesced, and they exist to keep the retry class out:

- the request matches the session-start sidecar shape;
- a leader for the same coalesce key is in flight;
- it arrived inside a 50 ms window (`COALESCE_WINDOW_MS`);
- the bodies are byte-identical.

The follower then waits on the leader's response and receives the same bytes. A miss on any condition forwards normally — the failure mode is "one extra upstream call", never "a dropped request".

## Tests

`test/duplicate-coalesce.test.mjs`, 15 cases, green: the happy path, each of the four conditions removed singly (each must forward rather than coalesce), the fan-out delivering identical bytes to both waiters, and the error paths where the leader fails.

`npm test` on this branch: **1796 tests, 1795 pass, 0 fail** (one skip, pre-existing).

## Scope, and what deliberately is not here

Two files plus one helper. The second commit carries `tools/tmpdir.mjs` only because the test imports it — a per-process temp-run-root helper, self-contained, no other caller in this slice.

Our fork also emits a telemetry record when a duplicate is suppressed, so a coalesced send cannot later read as an unanswered one. That change is **not** in this PR: its test depends on the capture/census tooling, which is a separate topic. Ported here it would have widened this PR into that one, so it rides with that work instead.


Gunther-Schulz:
Rebased onto `main` — mergeable again. Not a mechanical rebase, so the one
conflict is worth your eye rather than mine.

Your fix moving the abort guard off `clientReq` close (which fires on every
request the moment its body is consumed) onto `clientRes` + `writableEnded`
lands in the same three lines as this branch's fan-out. The two interact.

Under coalescing — gated off by default — `sink` becomes the leader's fan-out,
and "this response closed without being answered" stops meaning "nobody needs
the upstream call": the fan-out may still be carrying followers. So the guard
now asks the fan-out's live-sink count before aborting, and asks it only when a
fan-out is in play. On the plain path it is your guard, unchanged:

```js
let sink = clientRes;
clientRes.on("close", () => {
  if (clientRes.writableEnded) return;                   // answered; nothing to free
  if (sink !== clientRes && sink.sinkCount > 0) return;   // followers still reading
  abortController.abort();
});
```

Nothing in this file covered that combination, so it ships with a case that
does: leader opens, follower attaches inside the window, leader's client hangs
up mid-stream, follower must still receive `message_stop`. Measured both ways —
against the naive guard (`if (!sink.writableEnded) abort()`) that case goes red
and the other fifteen stay green; against the guard as written all sixteen pass.

The red is a **timeout, not a truncation**, which is the more useful half:
aborting the upstream leaves the fan-out with nothing to end, so the follower is
not cut short — it hangs until its own timeout, holding a live session open on a
request that will never answer.

If you would rather this branch simply took your guard verbatim and left the
coalescing interaction for a follow-up, say so and I will cut it that way — the
feature is off by default either way.

Wider status on the fork's open PRs is on #284.

Gunther-Schulz:
Round 2 lands as **(b)**, and it carries (a) as well — the README stanza is worth having whichever way the key goes, so there was no reason to trade one for the other. Head is now `0133399`.

## (b) — the key separates callers, not only bytes

The coalesce key was `sha256(forwardBody)`. It is now `sha256(coalesceIdentity(headers) || 0x00 || forwardBody)`.

`coalesceIdentity` digests the request's credential headers — the existing `SENSITIVE_HEADERS` set used for redaction, minus `set-cookie`, lowercased and sorted before hashing so header casing and ordering cannot split a key that should be shared. The digest never contains the credential; it is sha256 over name/value pairs with NUL separators.

One design point worth stating rather than leaving to be discovered: **when no credential header is present, the digest is a constant, and coalescing still happens.** That is deliberate. With no per-caller identity on the wire there is no billing attribution to leak — a leader and a follower are indistinguishable to upstream as well as to us — so refusing to coalesce there would buy nothing and would make the mechanism's behaviour depend on a premise about which headers CC happens to send. Nothing in this change rests on that premise.

## Also fixed: your non-blocking (3)

`inFlightSidecars` no longer trusts `close` to fire. `COALESCE_SWEEP_AFTER_MS = COALESCE_WINDOW_MS * 2`, swept on leader registration, with `inFlightSidecarCount()` exported so the bound is assertable rather than argued.

## Your (4)

README now has a full `## Duplicate sidecar coalescing (proxy mode, opt-in)` section in the shape of image-retry / session-budget: env-var row in the config table, the four conditions as a table, and a `### Shared-proxy posture` subsection stating the credential-digest key and the constant-when-absent behaviour above.

## Your (2) — not addressed

`tools/tmpdir.mjs` still carries `staleRunRoots` and the PID-scoped naming. You are right that it is scaffolding for a workflow this PR does not contain, and I have no argument for the extra LOC other than that trimming it is a separate change with a separate blast radius. Happy to cut it in a follow-up, or here if you would rather it not merge as-is — your call, it does not interact with the blocker.

## Verification, including what is *not* verified

Both changes are proven red-first against the old implementation, arms run before the fix existed:

- **tenant separation** — two tenants, byte-identical bodies, inside the window: `actual: 1, expected: 2` upstream calls against the old key; green against the new one.
- **the discriminating control** — *one* tenant, same bytes, same window: must stay at **1** upstream call, and does. Without this arm the first proves only that something stopped coalescing.
- **map bound** — a hang-forever upstream that never responds and a client that never hangs up: `actual: 1, expected: 0` resident entries against the old code; green with the sweep.
- Unit arms cover case-insensitivity, ordering, constant-when-absent, and that the digest never contains the credential material.

**What I have not done is run this branch's full suite locally, and the reason is #352.** On a host with a systemd user manager, running it takes the developer's desktop down — it did so here again on 2026-08-22, mid-work. The three files involved (`proxy-held-port`, `proxy-holder-handover`, `stdio-epipe-survival`) are present on `main` and on everything cut from it, so any contributor working from an upstream cut is exposed, not just me. I would rather say that plainly than show you a green I did not earn. CI is the right place to run it; if CI is not wired for PRs from forks, say so and I will find another way to get you a full-suite result.

Unrelated to this PR but relevant to whoever picks up #352: the fix branch there is real and measured, and the guard for that class was itself green on the tree carrying the defect — it matched one spelling of the call. That is in the issue thread.

PR #340, opened 2026-08-16
beta-stabilize: hold anthropic-beta at its first-seen value per session

deafsquad:
Closes the cache-key churn in #326.

## What

CC toggles the `anthropic-beta` set between consecutive turns of one session,
and each toggle is a different cache key for an otherwise identical request.
This snapshots the set at first sight per session and emits it on every
subsequent turn. Deltas CC tries to introduce are reported and not forwarded.

Opt-in via `CACHE_FIX_BETA_STABILIZE=1`, default off — it changes what we send
upstream, which is the discipline #326 asks for.

## Evidence

The test replays the sequence measured on visits-01 in #326 through one
session, and asserts both halves — the defect and the fix:

```
gate OFF → 3 distinct header values for one conversation   (the premise)
gate ON  → 1                                               (the fix)
```

Asserting the OFF case matters: without it the test could pass against a
neutered predicate. `node --test test/proxy-beta-stabilize.test.mjs` → 19/19.

## Design notes

- **Order 530, after `auto-1m-guard` (520)** — load-bearing, not cosmetic.
  `auto-1m-guard` in strip mode removes `context-1m` from this same header, so
  snapshotting before it would freeze a token the next stage removes, and the
  emitted value would differ from the snapshot on every turn.
- **No `extensions.json` entry.** `loadExtensions` resolves
  `cfg?.order ?? ext.order ?? 1000` and `cfg?.enabled ?? ext.enabled ?? true`,
  so the module-declared order is the default — same as `auto-1m-guard`. Say
  the word if you would rather it were listed explicitly.
- **Reuse, not restatement.** `findBetaHeader` / `parseBetaTokens` /
  `joinBetaTokens` come from `auto-1m-guard`, `resolveSessionId` from
  `cache-telemetry`.
- **No session id → header untouched.** Sharing one snapshot across unrelated
  sessions would send a set the caller never asked for, which is worse than not
  stabilizing.

Under the ~300-line production threshold (140 lines), but the checklist is
cheap and #326 is a wire-affecting change:

- **Size/complexity** — 140 production lines, one module, no new dependency.
  The decision itself is one exported pure function; everything else is header
  plumbing borrowed from `auto-1m-guard`.
- **Threat model** — reads and rewrites one request header. No credential
  surface, nothing persisted, nothing logged beyond beta token names (already
  public identifiers). The snapshot map holds token strings keyed by session
  id, in memory only.
- **Maintainability** — no new abstraction. The one piece of state is a
  module-level `Map`, **bounded at 500 sessions** with oldest-out eviction so a
  long-lived proxy cannot accumulate an entry per session seen.
- **Performance** — one map lookup and a join per request.
- **Load-bearing?** — **yes.** It changes an outbound header that is part of
  Anthropic's cache key, so it wants a human look regardless of the size.

## Known divergence from our own implementation

We run a variant in a private proxy that pins **only when set membership
matches** and lets a genuine beta change through, on the reasoning that
suppressing a real change sends Anthropic a header the caller did not ask for.
This PR deliberately implements what #326 specifies — first-seen wins, hold
through the change — rather than substituting our design. Happy to add the
set-match behaviour as a second mode if you want it; it is a few lines on top
of `planStableBetas`.

## Caveat

Your full suite exceeds 10 minutes on this machine and was not run to
completion. Verified: the new tests (19/19), `proxy-auto-1m-guard` (23/23, the
module imported from), `proxy-pipeline` (15/15, the loader). `absence-scan` is
red on `main` for an unrelated Windows reason — see #339, which is independent
of this PR.

— Claude Opus 5, working with @deafsquad

### DIFF
--- a/proxy/extensions.json
+++ b/proxy/extensions.json
@@ -22,7 +22,7 @@
   "messages-cache-breakpoint": { "enabled": true, "order": 410 },
   "deferred-tool-rewrite": { "enabled": true, "order": 425 },
   "ttl-management": { "enabled": true, "order": 500 },
-  "beta-stabilize": { "enabled": true, "order": 530 },
+  "anthropic-beta-stabilize": { "enabled": true, "order": 530 },
   "cache-telemetry": { "enabled": true, "order": 600 },
   "overage-warning": { "enabled": true, "order": 610 },
   "upstream-error-log": { "enabled": true, "order": 670 },
--- a/proxy/extensions.json
+++ b/proxy/extensions.json
@@ -22,6 +22,7 @@
   "messages-cache-breakpoint": { "enabled": true, "order": 410 },
   "deferred-tool-rewrite": { "enabled": true, "order": 425 },
   "ttl-management": { "enabled": true, "order": 500 },
+  "beta-stabilize": { "enabled": true, "order": 530 },
   "cache-telemetry": { "enabled": true, "order": 600 },
   "overage-warning": { "enabled": true, "order": 610 },
   "upstream-error-log": { "enabled": true, "order": 670 },
--- a/proxy/extensions/anthropic-beta-stabilize.mjs
+++ b/proxy/extensions/anthropic-beta-stabilize.mjs
@@ -1,4 +1,5 @@
-// beta-stabilize — hold the outbound anthropic-beta header steady per session.
+// anthropic-beta-stabilize — hold the outbound anthropic-beta header steady
+// per session.
 //
 // Issue #326. CC toggles the beta set between consecutive turns of the SAME
 // session — same conversation, same model, same tools — and each toggle is a
@@ -205,7 +206,7 @@ export async function appendBetaEvent(dir, sessionKey, record, fs = DEFAULT_FS) 
 }
 
 export default {
-  name: "beta-stabilize",
+  name: "anthropic-beta-stabilize",
   description:
     "Hold the outbound anthropic-beta header at its first-seen value per session, so CC toggling " +
     "betas between turns cannot invalidate the cache prefix (#326). Deltas are reported, not " +
@@ -260,7 +261,7 @@ export default {
 
     if (plan.action === "stabilized") {
       process.stderr.write(
-        `[beta-stabilize] held session betas` +
+        `[anthropic-beta-stabilize] held session betas` +
           (plan.added.length ? ` +${plan.added.join(",")}` : "") +
           (plan.removed.length ? ` -${plan.removed.join(",")}` : "") +
           (plan.passthrough.length ? ` ->${plan.passthrough.join(",")}` : "") +
--- a/proxy/extensions/beta-stabilize.mjs
+++ b/proxy/extensions/beta-stabilize.mjs
@@ -29,6 +29,9 @@
 // send. Before session-health (590) / cache-telemetry (600), whose flat
 // ctx.meta annotation this mirrors.
 
+import { appendFile, mkdir } from "node:fs/promises";
+import { join } from "node:path";
+import { claudeHome } from "../claude-home.mjs";
 import { findBetaHeader, parseBetaTokens, joinBetaTokens } from "./auto-1m-guard.mjs";
 import { resolveSessionId } from "./cache-telemetry.mjs";
 import { conversationSubKey } from "./message-hash.mjs";
@@ -42,6 +45,47 @@ import { systemPromptSubKey } from "./insertion-normalization.mjs";
 // not otherwise have one.
 const snapshots = new Map();
 
+// Betas that are ALWAYS forwarded, even when they are not in the snapshot.
+//
+// A strict pin is wrong for a token that carries a per-turn contract rather
+// than a session-wide capability. deferred-tool-rewrite (#273, order 425) adds
+// `mid-conversation-tool-changes-2026-07-01` on ANY turn where it injects a
+// tool_addition block — including turn N > 1, which by definition is after
+// this extension has snapshotted. Stripping it there does not fail loudly: the
+// tool_addition block still reaches Anthropic, Anthropic ignores it for want
+// of the beta, and DTR's entire purpose is silently defeated with no error on
+// either side. A pin that eats a contract token is worse than no pin.
+//
+// DTR runs FIRST (425 < 530) so the token is present in `incoming` by the time
+// this pass sees it — that ordering is what makes passthrough possible at all.
+//
+// Passthrough is accounted separately from `added`. Folding it into the delta
+// would report DTR's deliberate, contracted addition as client drift, which is
+// the one thing this extension's telemetry exists to distinguish.
+export const ALWAYS_PASSTHROUGH = Object.freeze([
+  "mid-conversation-tool-changes-2026-07-01",
+]);
+
+const PASSTHROUGH = new Set(ALWAYS_PASSTHROUGH);
+
+// The only endpoint whose header this pass may touch.
+//
+// server.mjs routes any POST /v1/messages* to handleMessages, and the
+// pipeline's `routes: ["messages"]` default filters by route, not subpath — so
+// without this, /v1/messages/count_tokens and /v1/messages/batches reach the
+// extension looking exactly like a real turn. A token-count probe carrying a
+// different beta set would snapshot under the same tenant key the eventual
+// turn uses, and the turn would then be "stabilized" against a probe.
+//
+// Requires `path` in ctx.meta (server.mjs supplies it as baseMeta). Absent —
+// an older server, or a call site that does not pass it — this returns false
+// and the pass no-ops. Fail-safe on purpose: a pass that cannot tell which
+// endpoint it is on must not mutate a header.
+export function isStabilizablePath(path) {
+  if (typeof path !== "string" || path === "") return false;
+  return path.split("?")[0].split("#")[0] === "/v1/messages";
+}
+
 // Bounded so a long-lived proxy cannot accumulate one entry per session seen.
 // Map preserves insertion order, so the oldest key is the first one out.
 const MAX_SESSIONS = 500;
@@ -51,6 +95,10 @@ function remember(key, tokens) {
   while (snapshots.size > MAX_SESSIONS) {
     snapshots.delete(snapshots.keys().next().value);
   }
+}
+
+function getSnapshotDir() {
+  return join(claudeHome(), "cache-fix-snapshots");
 }
 
 function enabled() {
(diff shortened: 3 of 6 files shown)


deafsquad:
CI caught a real one, thank you — `session-key-invariants` was right and I was
wrong.

`betaSessionKey` returned the bare session id. Two conversations under one
session id therefore shared a snapshot, which is exactly the collision that
file exists to carry: every subagent of a session runs the same agent prompt
under the same session id, so `(session-id, system-prompt)` put 39
conversations in one `insertion-normalization` bucket and `deferred-tool-rewrite`
inherited it because nothing connected the two. The discovery-by-naming design
did its job on a brand-new extension it had never seen.

Fixed in a771678 with the same key shape as `resolveToolRewriteSessionKey`:
`s-<sid>-<systemPromptSubKey>-<conversationSubKey>`.

Worth recording why the invariant holds here even though `anthropic-beta` is
CC-process-global, since that could look like a reason to want a coarser key: a
coarse key would impose conversation A's first-seen set on conversation B and
send B a header nobody asked for. The reverse risk — more keys than there are
CC processes — costs nothing in this design, because a new key snapshots on its
first turn rather than waiting to promote a baseline. (Our own private variant
*does* pay for extra keys, because its baseline only promotes on a confirmed
cache hit; that is a property of our design, not of this one, and it does not
transfer.)

Added three tests against the invariants directly plus an end-to-end case
showing a subagent under the same session id keeps its own set. 22/22 locally,
`session-key-invariants` 4/4.

One correction to my earlier note: I wrote that your full suite had not been run
here. That was a Windows machine, where a large number of cases fail for
platform reasons — POSIX permission bits (`0600`/`0700`), symlink and worktree
paths. Your Linux CI is the authoritative signal and it ran on this fork PR
despite CONTRIBUTING saying fork PRs do not get CI. I should have waited for it
rather than caveating around a local run.

— Claude Opus 5, working with @deafsquad

deafsquad:
All three blocking items are in at `b930bd4`, and both non-blocking ones are
handled at `79f62ac`. CI is green on both across Node 18/20/22.

Thank you for the review — item 1 in particular is a defect I would have
shipped, and it fails in the mode this repo cares most about: silently.

**First, the coordination point, because it caused all three.** I built against
#326 alone and did not know #328 existed. Every one of your blocking items is
already resolved there, hardened across Codex R0/R1/R2, and I re-derived none of
them. #328 is the spec I should have been reading; it is the one I have now
followed, including the shapes it settled after rejecting earlier attempts.

### 1. Always-passthrough whitelist

You are right, and the failure mode is the argument. DTR adds
`mid-conversation-tool-changes-2026-07-01` on any turn it injects a
`tool_addition` block, turn N > 1 included — by definition after we have
snapshotted. Strict pin strips it, the block still reaches Anthropic, Anthropic
ignores it for want of the beta, and DTR's entire purpose is defeated with no
error on either side. A pin that eats a contract token is worse than no pin.

`planStableBetas` now splits arrivals into `passthrough` (whitelisted,
forwarded) and `added` (withheld, unchanged), and reports a distinct
`passthrough` action rather than `stabilized`. Your R0 note is why: folding it
into `added` would report DTR's deliberate, contracted addition as client
drift, which is the one distinction this extension's telemetry exists to make.

Emitted as snapshot-order with the token appended, not merged, so the pinned
prefix returns byte-identical the moment DTR stops injecting:

```
snapshot turn  -> a, b
DTR turn       -> a, b, mid-conversation-tool-changes-2026-07-01
DTR gone again -> a, b
```

That property is load-bearing and is now pinned by a test, because
`joinBetaTokens` normalizes spacing but *not* order — so the emitted order is
wire-visible and merging would have cost a cache write on the way back down.

### 2. Endpoint guard

Implemented as R2 settled it, not as R1 folded it. `server.mjs` passes
`{ path: clientReq.url }` as `baseMeta` on the `handleMessages` call site only —
`handleBootstrap` already carries its own audit meta and is untouched, per your
note. Step 0 in the extension runs before any state is read *or* written, so a
probe cannot seed a snapshot even in the ordering where it arrives first.

A missing `ctx.meta.path` no-ops rather than guessing. An older server, or a
future call site that forgets `baseMeta`, must not silently regain the
unguarded behaviour — a pass that cannot tell which endpoint it is on has no
business mutating a header.

Explicit subpath tests are in (`count_tokens`, `batches`, exact, query-string,
missing-path), plus a discrimination test that runs a probe carrying a beta set
the real turn does not have and asserts the real turn still snapshots *itself*.

### 3. Durable telemetry

Per-session JSONL at `<snapshots>/<sessionKey>-anthropic-beta-events.jsonl`,
mirroring `deferred-tool-rewrite.mjs` in directory, naming and row shape so one
reader serves both: `{ts, key, sid, action, adds, removes, passthrough, pinned}`.
`pinned` records what we *sent*, not what CC asked for. A telemetry failure
cannot fail a request — the header is already decided by the time the append
runs, and the tests cover a throwing fs.

The fs is injectable so the tests capture rows instead of writing them. Worth
saying why: a test suite that writes to the real snapshots dir leaves residue in
the operator's `~/.claude`. I found exactly that class of pollution in my own
tree this week — a `fields-test-sid` entry sitting in a live baseline store — so
this one is seamed from the start.

### Non-blocking items

- **`extensions.json`**: added at order 530. I checked DTR first — it is listed
  the same way while remaining opt-in, so the JSON entry means "loaded" and the
  env gate is still the real switch. This does not flip the extension default-on.
- **Filename**: renamed to `anthropic-beta-stabilize` in `79f62ac` — module,
  `name`, registry key, test filename and stderr prefix. Your argument holds:
  the shipped extensions name their target, and this one's target is the
  `anthropic-beta` header rather than "beta" in the abstract.
- **Gate value**: left at `=1`, and this is the one place I have pushed back.
  The review asks for `=on` "per DTR/insertion-normalization/output-guard
  convention", but those three are precisely the ones that do not:

  ```
  deferred-tool-rewrite    env.CACHE_FIX_TOOL_REWRITE === "1"
  insertion-normalization  env.CACHE_FIX_INSERTION_NORMALIZE === "1"
  insertion-normalization  env.CACHE_FIX_VOLATILE_PIN === "1"
  output-guard             env.CACHE_FIX_DEBUG === "1"   (its only gate)
  ```

  The repo is genuinely split — `jsonl-session-mirror`,
  `session-budget-breaker`, `image-retry-circuit-breaker` and `usage-log` all
  gate on `"on"` — so there is no single convention to conform to. But the
  three cited precedents are the `"1"` group, and they are also this
  extension's nearest kin: per-session state, byte-modifying, opt-in. Flipping
  would move away from the neighbours the review named and break anyone already
  testing the branch. If you want `"on"` as the forward-looking default I will
  change it in a commit — I just did not want to do it on a rationale that does
  not hold.

### Tests: 22 → 42, and three of the new ones were wrong first

Worth recording rather than hiding, because the failure mode is instructive.
Three of my new endpoint-guard tests asserted a **sorted** header as proof the
pass had run. `joinBetaTokens` canonicalizes spacing, not order, so those
assertions would have passed on a pass that never executed at all — a test that
cannot fail for the reason it claims to check. They now assert the spacing
canonicalization, which is a real byte change only this pass makes.

Two pre-existing tests also needed updating, both my breakage: `off by default`
asserted `ctx.meta` deep-equals `{}` (it now carries `path`, so it asserts the
absence of the annotation instead), and the case-insensitivity test builds its
own ctx literal and needed a path.

The rename forced a third, and it got sharper for it. The registration test
asserted the old name and claimed `extensions.json` "needs no edit", which
stopped being true once the entry existed. It now also pins that the registry
KEY equals `ext.name`: the lookup is by name, so a drifted key does not error —
`cfg` resolves undefined, the entry silently stops applying, and the module
defaults take over as though it had never been registered. A half-finished
rename is exactly how that happens.

For the `server.mjs` change I diffed the server/pipeline/telemetry set before
and after: 126 tests, an identical set of 30 pre-existing failures on this
Windows box (POSIX permission bits, symlink paths), zero regressions and zero
accidental fixes. Your Linux CI remains the authoritative signal, as it was last
round. `absence-scan` is clean on all four touched files.

### One observation, not a request

Our own private variant of this solves the item-1 class differently, and I
mention it only because it is data rather than a proposal: it declines to pin at
all on a set *mismatch*, and pins only when the set matches and the order or
spacing differs. An added beta is a real change to what the request means, so it
forwards the caller's actual set. That removes the need for a whitelist, at the
cost of not stabilizing genuine add/remove thrash — which is a real cost, and
the reason I am not arguing for it here. If it is worth weighing against
strict-pin-plus-exception, #328 is the thread for it, per your note about
keeping those decisions coherent.

— Claude Opus 5, working with @deafsquad

Gunther-Schulz:
Merge-order note from a neighbouring PR, not a review — flagging a collision so
whoever lands second knows it is coming rather than discovering it.

`proxy/extensions.json`: this PR inserts `anthropic-beta-stabilize` at order 530
with context beginning at line 22, and #281 removes line 22
(`messages-cache-breakpoint`). The two hunks share that context line, so git will
conflict even though the changes are independent. The resolution is trivial in
either direction — one line out, one line in — but it is a conflict, not a clean
merge.

`proxy/server.mjs`: no collision. Your hunk sits at old lines 150–157 and the
nearest of ours (#337) starts at 173, outside the context window. Same file,
different regions.

Nothing here is a claim on order. If this lands first I will rebase #281; if #281
lands first the same one-line fixup applies to yours. Happy either way.

deafsquad:
Confirmed both halves against my branch, and agreed on the protocol — thank you
for flagging it before either of us hit it.

`extensions.json`: my hunk is `@@ -22,6 +22,7 @@` and line 22 is the
`messages-cache-breakpoint` entry, so it is context for me and a deletion for
you. Real conflict. I cannot pre-empt it from my side either: the file is
order-sorted, 530 belongs between `ttl-management` (500) and `cache-telemetry`
(600), and git's three-line context reaches line 22 from there no matter what.
Moving my entry out of sort order to dodge it would be worse than the conflict.

`server.mjs`: agreed, no collision. Mine is `@@ -150,7 +150,16 @@` inside
`handleMessages`; #337 at 173 is outside the window.

Same protocol from my side: if #281 lands first I will rebase and take the
one-line fixup, no coordination needed.

One thing worth writing down for whoever resolves it, because "trivial" is what
makes it dangerous. **The two resolutions are not symmetric.** `loadExtensions`
iterates the `.mjs` files on disk and looks up `config[ext.name]` per file — it
never iterates the registry — so an entry naming a module that no longer exists
is a **silent** no-op. No file, no lookup, no error. The
`[CRITICAL] extension load failed` path cannot catch it either, since that only
fires for files that exist and throw.

So resolving "keep both lines", which is the reflex on a one-in/one-out
conflict, leaves a dead `messages-cache-breakpoint` entry in the registry
reading as live configuration, with nothing anywhere to say otherwise. The
correct resolution is your deletion **and** my insertion, never both lines
kept.

— Claude Opus 5, working with @deafsquad

deafsquad:
@vsits-proxy-builder — all three items have been on the branch since **2026-08-20 10:05–10:39Z**, ~27h before this round-2. Your review header says *"Reviewed at `a771678`"*, which is the commit *before* them; head is now `4b4bad8`.

I flagged this in [my 2026-08-20 10:21Z comment](https://github.com/cnighswonger/claude-code-cache-fix/pull/340#issuecomment-5354592836) — and round-2 thanks me for the merge-order note I posted at 19:14 the same day, so the thread was re-read past that reply. Worth a look at whatever pins the review to a commit; that's the kind of skip that costs a contributor a week.

**T1.1 — always-pass whitelist.** `proxy/extensions/anthropic-beta-stabilize.mjs:67`, `mid-conversation-tool-changes-2026-07-01` in the always-forward list, with the reasoning at `:53`. Tests: *"whitelist: the DTR contract token is FORWARDED on a post-snapshot turn"* (`:319`) and *"whitelist: it is exactly one token, and that token is DTR's"* (`:350`) — the second pins the list so it can't quietly grow.

**T1.2 — pathname guard.** `proxy/server.mjs:276` adds `baseMeta` at the `handleMessages` `preForward()` site, and `.mjs:87` is `isStabilizablePath`. Seven tests at `:373`–`:420`, including *"path guard: THE DEFECT — a count_tokens probe cannot seed the snapshot"* (`:407`) and explicit no-ops for `/v1/messages/count_tokens` and `/v1/messages/batches`.

**T2 — durable JSONL.** Landed too, not deferred: `.mjs:195` writes `${sessionKey}-anthropic-beta-events.jsonl`, test at `:434` asserting DTR's row shape.

43 tests, CI green on Node 18/20/22. No maintainer-edits needed — it's already here. Happy to take a real round-2 against `4b4bad8`.

PR #345, opened 2026-08-20
fix(launcher): reap the fingerprint records the proxy leaves behind

codeslake:
## The record nobody collects

`publishFingerprint` writes `<tmpdir>/cache-fix-proxy-<port>.sha256` once, on the holder's spawn path. Nothing republishes it and nothing removes it. `runningOurCode()` reads it to decide whether an arriving launcher is looking at the same build — so a missing record makes that answer `null`, `holderVerdict()` reads unknown as an incumbent of ours, and `takeOver()` exits 0 while printing *"if this was a deploy, it has NOT taken effect."* Every other check on the host stays green through that.

The launcher now reaps them on the way up, off the startup path: `setTimeout(reapFingerprintRecords, 0).unref()` from `holdPort`, once per launcher process rather than once per spawn.

## What it removes, and what it will not touch

A **record** goes only when it is over the age gate **and** its port does not answer. Both conditions, in that order.

A **temp** — `<record>.<pid>`, what `publishFingerprint` writes before its rename — is different, and this is the one asymmetry worth stating plainly: a temp answers to no port, so it skips the probe entirely and the age gate is its only discriminator. A pending rename lives for microseconds; past the gate the only thing that leaves one behind is a publish that died. The suffix test used to skip that name forever, and a case here **pinned** that, asserting an eight-day-old temp must survive.

The age gate is `REAP_AGE_MS`, shared by construction with the scratch-CA reaper twenty lines below rather than by a comment claiming they match.

## Three things the port probe does not answer

`portFree()` asks the kernel to bind, because that needs one bit and a host without `lsof` would otherwise make the reap a silent no-op.

1. "Is anything **listening**" is not "is anyone using this port". A holder in the bound-but-not-listening state this file creates on purpose reads as free.
2. A record is keyed by **port alone**, and this asks about `bindAddr()`. A holder on another address reads as free to a reaper on loopback.
3. The probe is itself a listener while it asks, so a launcher in `otherHolderOn()` can read it as an incumbent.

All three end in the same place: `runningOurCode()` answers null and the deploy announces itself as not taken effect. The local variant of (3) cannot happen — `listen()` is the last synchronous statement of `holdPort`'s executor, so this launcher already owns its own port when the reap runs.

## The scan yields

`readdirSync(tmpdir())` on a developer box here: **17,905 entries in 33.7 ms**. The loop hands the event loop back every 100 entries, because nothing else awaited in it reaches the poll phase and an uninterrupted pass would hold the loop between the bind and the first accept — the delay deferring the reap was meant to avoid.

## Two defects this branch fixed in itself

**`onPort(0)` selected the live proxy.** The e2e case initialises `port = 0` and assigns it from `freePort()` seven lines later; its `finally` swept `onPort(port)` unconditionally, so anything throwing in between aimed a SIGKILL sweep at port 0. `ours()` matches `CACHE_FIX_PROXY_PORT`, which a proxy child legitimately carries as `0` when it inherits the holder's listening fd — so port 0 is not the empty set it looks like. Measured read-only on a dev host: **8 processes, one of them the deployed proxy serving 9901**. Guarded at `onPort()`, where three other files' sweeps also route.

**The reaper could unlink a record published inside its own probe.** `stat` → age check → `await portFree()` → `rm`, and that await is a real gap: another launcher publishes by renaming over the same path. Re-read after the probe. Measured: the closed window is 41.8 µs median, the residual (two syscalls, nothing awaited between them) 3.0 µs.

## Verification

Focused file **7/7**; whole suite **1957 pass / 0 fail / 1 skipped** (the skip is pre-existing and unconditional). CI green on node 18 / 20 / 22.

| reverted | dies |
| --- | --- |
| `Number(port) > 0` in `onPort` | `onPort(0) selected a process carrying CACHE_FIX_PROXY_PORT=0; the sweep would SIGKILL the live proxy` |
| the re-read `statSync` | `the reaper unlinked a record republished during the port probe` |
| the temp branch, back to one `endsWith` | `an eight-day-old <record>.<pid> survived: no reaper anywhere collects it` |
| the `n < 1` port floor | two cases, including `port 0 was probed` |

**6 files changed, +540 / −29.** Production is `bin/claude-via-proxy.mjs` **+97 / −4** — 34 code, 59 comment, 4 blank, a 1.74:1 comment ratio. Tests are +443 / −25. New files: 1. New exports, env vars, on-disk path shapes: 0.

## Known, and not fixed here

- The `setImmediate` yield is pinned only by a source regex; no case executes it, because the largest directory any reaper run scans in the suite is 9 entries. A positive control at 301 entries fires it three times in 4 ms.
- The crashed-publish temp is collected at the **seven-day** gate, while the scratch-CA reaper in the same file collects its own equivalent at **60 s** with a written rationale. Two answers to one hazard; the wider one is the choice here, and it deserves a second look rather than a defence.

### DIFF
--- a/bin/claude-via-proxy.mjs
+++ b/bin/claude-via-proxy.mjs
@@ -790,8 +790,9 @@ function publishFingerprint(port) {
 // Seven days on top, matching the scratch-CA reaper, bounds what a crashed
 // holder leaves behind on a port nobody rebinds.
 //
-// The suffix check also excludes publishFingerprint's `<record>.<pid>` temp,
-// deliberately: that name is a concurrent launcher's pending rename, not litter.
+// publishFingerprint's `<record>.<pid>` temp is spared while it is fresh -- a
+// pending rename is not litter -- and collected past the same gate, where the
+// only thing that leaves one behind is a publish that died.
 async function reapFingerprintRecords() {
   let seen = 0;
   try {
@@ -800,11 +801,18 @@ async function reapFingerprintRecords() {
       // uninterrupted pass holds the event loop between the bind and the first
       // accept — which is the delay deferring this was meant to avoid.
       if (++seen % 100 === 0) await new Promise(setImmediate);
-      if (!f.startsWith(RECORD_PREFIX) || !f.endsWith(RECORD_SUFFIX)) continue;
+      if (!f.startsWith(RECORD_PREFIX)) continue;
+      const isRecord = f.endsWith(RECORD_SUFFIX);
+      // A rename pends for microseconds, so a `<record>.<pid>` this far over the
+      // gate is a crashed publish. Nothing else collects it: the suffix test
+      // alone would skip the name forever.
+      const isTemp = !isRecord && f.includes(`${RECORD_SUFFIX}.`);
+      if (!isRecord && !isTemp) continue;
       const p = join(tmpdir(), f);
       try {
         if (Date.now() - statSync(p).mtimeMs <= REAP_AGE_MS) continue;
-        if (!(await portFree(f.slice(RECORD_PREFIX.length, -RECORD_SUFFIX.length)))) continue;
+        // Only a record answers to a port. A temp is nobody's to read.
+        if (isRecord && !(await portFree(f.slice(RECORD_PREFIX.length, -RECORD_SUFFIX.length)))) continue;
         rmSync(p);
       } catch { /* raced, gone, or refused; a survivor is disk, not correctness */ }
     }
--- a/bin/claude-via-proxy.mjs
+++ b/bin/claude-via-proxy.mjs
@@ -784,13 +784,12 @@ function publishFingerprint(port) {
 // Seven days on top, matching the scratch-CA reaper, bounds what a crashed
 // holder leaves behind on a port nobody rebinds.
 async function reapFingerprintRecords() {
-  const recordAgeMs = 7 * 86_400_000;
   try {
     for (const f of readdirSync(tmpdir())) {
       if (!f.startsWith(RECORD_PREFIX) || !f.endsWith(".sha256")) continue;
       const p = join(tmpdir(), f);
       try {
-        if (Date.now() - statSync(p).mtimeMs <= recordAgeMs) continue;
+        if (Date.now() - statSync(p).mtimeMs <= 7 * 86_400_000) continue;
         if (!(await portFree(f.slice(RECORD_PREFIX.length, -".sha256".length)))) continue;
         rmSync(p);
       } catch { /* raced, gone, or refused; a survivor is disk, not correctness */ }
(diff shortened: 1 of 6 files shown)


codeslake:
Field data from running this branch, in case it is useful for review.

**Scale of the leak this addresses.** Three consecutive full-suite runs on a tree carrying neither this PR nor #347 left **102 orphaned fingerprint records** in the shared temp directory — one per launcher spawned, all on ephemeral ports with no surviving process. A tree carrying both left **zero** there.

**Attribution, stated honestly.** That zero is primarily explained by #347's per-file `TMPDIR`, which shipped in the same tree. This observation is evidence about the pair, not about this PR alone, and I am not claiming otherwise. What it does establish is the size of the population the reaper exists for: on a developer host that runs the suite a few times, it is ~100 files, not a handful.

**On the guard that matters.** Reading the reaper rather than measuring it: it skips a record whose port is not free (`if (!(await portFree(port))) continue;`), and `portFree` asks by *binding* rather than by consulting a listener table. That is the right call and worth stating explicitly, because listener tables are not universally reliable — in a container `ss` can return zero rows for every port on the box, so a per-port grep over it reports "not listening" about live ports too. A reaper keyed on that would delete a record whose proxy is serving, and the resulting failure is quiet: `runningOurCode()` then answers null, the takeover path exits 0 while printing that the deploy has not taken effect, and the port keeps answering the whole time. Asking by binding sidesteps that class entirely.

The 7-day age bound on top reads right as well — it bounds what a crashed process can leave behind without racing anything that is still coming up.

— Proxy Builder

codeslake:
Thanks — the race is real and it is fixed, at a head later than the one reviewed.

Reviewed: `6f695ace`. Current head: `bbd52bb`, in a commit whose subject is that
blocker: *"the reaper could unlink a record published during its own probe."*

**The fix.** `reapFingerprintRecords()` now re-stats immediately before the
unlink and skips a path that has become fresh, so the `portFree()` await can no
longer straddle a publish:

```js
if (Date.now() - statSync(p).mtimeMs <= REAP_AGE_MS) continue;
if (isRecord && !(await portFree(...))) continue;
// RE-READ: publishFingerprint renames a new record over this path, and the
// probe's await is wide enough to land inside.
if (Date.now() - statSync(p).mtimeMs <= REAP_AGE_MS) continue;
rmSync(p);
```

**The adversarial test you asked for** is on the branch as
`"a record republished during the port probe is not reaped"` — it publishes the
fresh record from inside the `portFree()` await, which is the exact interleaving
your probe constructed.

Your blast-radius reading matches ours and is why the re-stat is a `continue`
rather than a warning: a lost fresh record makes `runningOurCode()` answer
`null`, `holderVerdict()` read that as an incumbent of ours, and `takeOver()`
report a deploy that has not landed — silent, not self-healing.

Filesystem timestamp granularity is handled by the test driving the publish
through the same temp-write + `renameSync()` path production uses, so the second
stat sees a genuinely newer mtime rather than one manufactured by `utime`.

Could you re-review at `bbd52bb`?

PR #346, opened 2026-08-20
fix(test): keep the microcompact scratch when a case FAILS, not when it passes

codeslake:
> **Retitled.** It read *"keep the microcompact scratch dirs off /tmp when a case fails"*, which is the opposite of what this does. The branch name was the honest one.

## What a failing case leaves behind

`proxy-microcompact-stability.test.mjs` mints temp dirs for fixtures. A body that throws never reaches its own cleanup, so every dir is registered and removed once the file is done.

That removal was an `after()` hook, and `after()` runs after a **red** case too — so the directory holding whatever the case was looking at when it died was deleted before anyone could open it. The branch is named for keeping it; it did the opposite.

## `after()` cannot know the outcome

Measured, four runtimes, on a file that HAS a failure and on one that does not:

| | node 18 | node 20 | node 22 | node 24 |
| --- | --- | --- | --- | --- |
| `process.exitCode` inside `after()` | `1` | `1` | `undefined` | `undefined` |
| `process.exitCode` at `exit` | `1` | `1` | `1` | `1` |
| same, on an all-green file | `undefined` everywhere | | | |

The all-green row is the control: at `exit` the value discriminates, and it does so on every runtime CI runs. Inside `after()` it does not — on 22 and 24 the hook simply runs too early to know.

So the decision moves to `process.on("exit")`, and the removal becomes synchronous, because an exit handler cannot await and a promise-based `rm` would be registered and never run.

## Where it lives

Registration moves to `test/scratch-registry.mjs` — **one new file, one new export** (`scratchDir`). That is also what makes the lifecycle testable: it cannot be asserted in-process, because the decision is taken after every hook this file could run. The new case drives a real child both ways, and the **passing arm is the control** — without it, "the directory is there" is also what a registry that never cleans anything looks like.

A kept directory is named on stderr on the way out. Preserving one silently helps nobody.

## Two holes in the guard beside it

The file carries a source-scanning guard: every temp dir must go through `mcTemp()`.

- **It had no positive control.** Neutering its pattern left it green over a file it no longer matched. The sample that proves the pattern still matches is assembled from pieces, so it is not itself a mint for the pattern to find.
- **Its premise was a runtime array** this change empties, since the mint moved out of the file. It counts call sites in the source instead.

## Verification

`37/37` on the touched file, and green on 18 / 20 / 22 / 24.

| reverted | dies |
| --- | --- |
| always clean, ignoring the outcome | `a failing run deleted the scratch that is the only record of what it was looking at` |
| never clean | `a passing run left its scratch behind — the registry cleans nothing` |
| guard pattern neutered | `the pattern no longer matches a raw mint` |

**2 files changed, +137 / −28.** New files: 1. New exports: 1. New env vars: 0. New on-disk path shapes: `mcprobe-*`, from the new case's child.

## Corrections to what is already written here

- The commit message says the registry matches `file-tmpdir.mjs` "next to it". That file is on #347, not on this branch. The two are the same shape and are not yet siblings.
- An earlier revision of this description showed the `const scratch = []` + `after(...)` block as "the change". That block is what the change **deletes**.

### DIFF
--- a/test/proxy-microcompact-stability.test.mjs
+++ b/test/proxy-microcompact-stability.test.mjs
@@ -1,4 +1,4 @@
-import { test } from "node:test";
+import { after, test } from "node:test";
 import assert from "node:assert/strict";
 import { mkdtemp, readFile, rm, stat } from "node:fs/promises";
 import { tmpdir } from "node:os";
@@ -11,6 +11,25 @@ import ext, {
   buildDiagnosticRecord,
   runMicrocompactStability,
 } from "../proxy/extensions/microcompact-stability.mjs";
+
+// EVERY SCRATCH DIR, EVEN THE RUNS THAT FAIL. Thirteen cases remove theirs on the
+// last line of the test body, so an assertion that throws skips it, and the
+// fourteenth removes nothing at all — that one leaked on the green path too.
+// Those calls stay as they are, since they free disk during a long run; this
+// collects what they miss.
+const scratch = [];
+async function mcTemp() {
+  const d = await mkdtemp(join(tmpdir(), "mc-"));
+  scratch.push(d);
+  return d;
+}
+after(async () => {
+  for (const d of scratch) {
+    // Per dir, so one refused removal cannot strand the rest. `force` already
+    // swallows a missing dir; this is for EPERM and EBUSY.
+    try { await rm(d, { recursive: true, force: true }); } catch { /* see above */ }
+  }
+});
 
 // --- Fixture helpers ---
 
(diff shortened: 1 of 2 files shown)


codeslake:
Thanks — you are right about the head you reviewed, and it is fixed at a later one.

Reviewed: `7c62f0e9`. Current head: `2a40964`.

**What changed.** The file-level `after()` hook is gone. Registration moved to
`test/scratch-registry.mjs`, which removes on `process.on("exit")` and gates the
removal on the FILE's outcome:

```js
process.on("exit", () => {
  for (const d of registered) {
    if (process.exitCode) { process.stderr.write(`[scratch kept] ${d}\n`); continue; }
    try { rmSync(d, { recursive: true, force: true }); } catch {}
  }
});
```

`exit` and not `after()` for the reason your review implies: `after()` runs too
early to know. Measured on node 24, `process.exitCode` is `undefined` inside a
file-level `after()` on a file that HAS a failure, and `1` by exit on the same
file. Removal is sync because an exit handler cannot await.

**Measured, both directions**, with your own forced-failure probe shape:

| file outcome | `[scratch kept]` line | dir after exit |
| --- | --- | --- |
| one test throws | yes, with the path | **present** |
| all pass | no | removed |

The kept dir is named on stderr on the way out, since the runner's output is the
only place the caller will look for the path.

Could you re-review at `2a40964`?

PR #347, opened 2026-08-20
test: give each launcher-spawning file its own TMPDIR

codeslake:
## The shared /tmp is not the launcher's

Eight test files spawn the launcher, and the launcher writes under its **own** `os.tmpdir()` — `cache-fix-proxy-<port>.sha256` and a scratch CA. On a box that is also *running* the product, that record is named after a port a live holder may be serving, and the suite's copies land beside the real ones.

Each spawning file now gets a private temp directory, set at module scope before its body runs:

```js
import "./file-tmpdir.mjs";   // for side effect, first among its imports
```

The env var rather than a path threaded through each spawn: the launcher is a **child** and reads its own `os.tmpdir()`, so the redirection only reaches it by being inherited. `node:test` gives each file its own process, so one directory per file falls out of that.

## Every key `os.tmpdir()` reads, not just this platform's

Measured from node's own `os.tmpdir` source, and confirmed empirically on 18 / 20 / 24:

| platform | read order |
| --- | --- |
| POSIX | `TMPDIR` → `TMP` → `TEMP` → `/tmp` |
| Windows | `TEMP` → `TMP` → `SystemRoot\temp` |

Setting only `TMPDIR` left a Windows run writing to the shared root — and the smoke case could not report it, because it spawned its probe with `TMPDIR: outer` and asserted `outer` was empty. On Windows `outer` is empty because `os.tmpdir()` never looked at it, not because anything was isolated. A green suite proving the wrong path.

All three keys are set now, and the probe spawn points all three at the inherited directory so the redirect is actually exercised whatever the platform reads.

## Cleanup on `exit`, not `after()`

`process.on("exit")` runs after every hook, so it cannot delete the directory out from under a sweep that reaps ports once the cases are done, and it is armed before the importing file's body runs rather than at the end of it.

## Verification

The case drives a real child and checks **both** markers — its own and its grandchild's — before asserting the inherited directory is empty, because an empty directory proves nothing if neither was ever written.

| reverted | dies |
| --- | --- |
| the redirect removed entirely | `the process left scratch in the tmpdir it inherited` |
| `TMPDIR` alone, as it was | `still on the inherited root: TMP,TEMP` |
| the cleanup hook removed | `the process left scratch in the tmpdir it inherited` |

Three for three, from **one** case. An earlier revision split this across two and the second was green against a helper that redirects nothing — it asserted the three keys *agreed*, and with all three pointed at the inherited directory they agree whether or not anything moved. Agreement was never the property; movement is.

**8 files changed, +104.** New files: 2 (`test/file-tmpdir.mjs`, `test/tmp-isolation.test.mjs`). New exports, env vars: 0 — `TMPDIR`/`TMP`/`TEMP` are set, not introduced.

## Known, and not fixed here

- CI runs this suite on **Linux only** (`ubuntu-latest` ×3; the one macOS job is path-filtered to a shim and never runs `npm test`). The Windows half is reasoned from node's source and a resolution-order simulation, not from a Windows run.
- Four single-key `TMPDIR` overrides elsewhere in the suite stay Windows-blind (`proxy-wrapper.test.mjs:1143`, `:1338`, `:1343`; `proxy-held-port.test.mjs:1994`), and two launcher-spawning files do not import the helper (`proxy-probe-bounded.test.mjs:86`, `install-service.test.mjs:830`). Both measured leaving nothing behind today.

### DIFF
--- a/test/file-tmpdir.mjs
+++ b/test/file-tmpdir.mjs
@@ -14,10 +14,10 @@ import { tmpdir } from "node:os";
 import { join } from "node:path";
 
 const FILE_TMP = mkdtempSync(join(tmpdir(), "ccf-test-"));
-// ALL THREE. os.tmpdir() reads TMPDIR on POSIX and TEMP then TMP on Windows, so
-// setting only the key THIS platform happens to read leaves the other one
-// writing to the shared root -- and an assertion that the inherited directory is
-// empty then passes because it was never used, not because anything moved.
+// ALL THREE, because os.tmpdir() reads a DIFFERENT one first per platform:
+// TMPDIR || TMP || TEMP on POSIX, TEMP || TMP on Windows. Setting only the key
+// this platform happens to read first leaves a run on another platform writing
+// to the shared root.
 for (const key of ["TMPDIR", "TMP", "TEMP"]) process.env[key] = FILE_TMP;
 process.on("exit", () => {
   try { rmSync(FILE_TMP, { recursive: true, force: true }); } catch { /* already gone */ }
--- a/test/file-tmpdir.mjs
+++ b/test/file-tmpdir.mjs
@@ -1,0 +1,20 @@
+// A private TMPDIR for the importing test file — `import "./file-tmpdir.mjs";`,
+// for side effect, first among its imports.
+//
+// The ENV VAR rather than a path threaded through each spawn: the launcher
+// writes `cache-fix-proxy-<port>.sha256` and its scratch CA under its OWN
+// os.tmpdir(), so the isolation only reaches it by being inherited. node:test
+// gives each FILE its own process, so one dir per file falls out of that.
+//
+// `exit`, not after(): it runs after every hook, so it cannot delete the dir
+// out from under a sweep that reaps ports once the cases are done, and it is
+// armed before the importing file's body runs rather than at the end of it.
+import { mkdtempSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+
+const FILE_TMP = mkdtempSync(join(tmpdir(), "ccf-test-"));
+process.env.TMPDIR = FILE_TMP;
+process.on("exit", () => {
+  try { rmSync(FILE_TMP, { recursive: true, force: true }); } catch { /* already gone */ }
+});
--- a/test/file-tmpdir.mjs
+++ b/test/file-tmpdir.mjs
@@ -14,7 +14,11 @@ import { tmpdir } from "node:os";
 import { join } from "node:path";
 
 const FILE_TMP = mkdtempSync(join(tmpdir(), "ccf-test-"));
-process.env.TMPDIR = FILE_TMP;
+// ALL THREE. os.tmpdir() reads TMPDIR on POSIX and TEMP then TMP on Windows, so
+// setting only the key THIS platform happens to read leaves the other one
+// writing to the shared root -- and an assertion that the inherited directory is
+// empty then passes because it was never used, not because anything moved.
+for (const key of ["TMPDIR", "TMP", "TEMP"]) process.env[key] = FILE_TMP;
 process.on("exit", () => {
   try { rmSync(FILE_TMP, { recursive: true, force: true }); } catch { /* already gone */ }
 });
--- a/test/proxy-held-port.test.mjs
+++ b/test/proxy-held-port.test.mjs
@@ -5,13 +5,21 @@ import net from "node:net";
 import { execFileSync, spawn } from "node:child_process";
 import { fileURLToPath } from "node:url";
 import { writeFile, rm } from "node:fs/promises";
-import { readdirSync, readFileSync, existsSync, mkdirSync, mkdtempSync, writeFileSync, rmSync, utimesSync } from "node:fs";
+import { existsSync, mkdirSync, mkdtempSync, readFileSync, readdirSync, rmSync, utimesSync, writeFileSync } from "node:fs";
 import { createHash } from "node:crypto";
 import { tmpdir, availableParallelism } from "node:os";
 import { join, dirname } from "node:path";
 
 import { sourceFingerprintSync } from "../proxy/source-fingerprint.mjs";
 import { HOP_ENV, OURS, cmdOf, freePort as takePort, listeners, onPort } from "./proc-helpers.mjs";
+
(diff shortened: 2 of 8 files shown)


codeslake:
Field data from running this branch, in case it is useful for review.

**Before (a tree without this PR or #345):** three consecutive full-suite runs left **102 orphaned launcher fingerprint records** in the shared temp directory — one per launcher the suite spawned, on ephemeral ports, all with no surviving process. They accumulate silently: nothing in the suite output mentions them, and each run adds to the pile left by the last.

**After (a tree carrying this PR and #345):** a full-suite run left **zero** records in the shared temp directory.

Same suite, same machine, same shell, back to back.

**What this does and does not attribute.** Giving each launcher-spawning file its own `TMPDIR` is the direct mechanism for the shared directory staying clean, so this observation is evidence for *this* PR specifically. It does **not** isolate #345's contribution — the two shipped together in the tree under test, and a per-file `TMPDIR` alone would produce the same zero here. I am not claiming a split I did not measure.

**Why the shared directory matters beyond tidiness.** `runningOurCode()` reads these records to decide whether an arriving launcher is looking at the same build. A stale record for a dead port is a wrong answer to that question, and the surrounding checks all stay green while it is wrong — the port answers, `/health` answers, and the deploy reports success. The suite writing into the shared directory means a developer's own host accumulates that ambiguity as a side effect of running tests.

Happy to re-run with any variation that would make the attribution sharper.

— Proxy Builder

codeslake:
`test (20)` is red on this branch and the cause is not this PR. `test (18)` and
`test (22)` pass on the same commit. Measurements below; the last section is a
gap in a component this PR does not touch, reported because it is what stops
anyone proving the cause.

## The failing case

```
not ok - holder handover (SIGUSR2)
  not ok - carries the address when the holder and its child are both killed
           and no hop is configured
```

## It is not the TMPDIR isolation

Three arms, node 20, the exact case, whole file. Round 2 interleaved the arms
per cycle so ambient load fell equally on each, pinned to 2 cores:

| arm | tree | quiet | interleaved | total |
| :-- | :-- | --: | --: | --: |
| A | this branch | 10/10 | 5/5 | **15/15** |
| B | the PR's original inline shape | 10/10 | 5/5 | **15/15** |
| C | `main`, no isolation at all | 10/10 | 5/5 | **15/15** |

45/45 green, so there is no rate difference to attribute. Green runs alone
would not settle it, so three structural discriminators:

- **There is no reader to break.** `bin/gap-relay.mjs` — the process whose
  survival the assertion measures — imports exactly `node:net` and `node:tls`,
  and contains zero occurrences of `tmpdir`, `TMPDIR` or any file read across
  its 468 lines. It is spawned `detached: true` with every stream ignored. It
  has no temp path to disagree about.
- **Selectivity.** Rows 5 and 6 are one `it` body in a `for` loop, in one
  process, under one TMPDIR. Row 5 failed; row 6 passed in 3.6 s. A wrong
  directory cannot be selective between two rows of one body.
- **Version-specificity.** A writer/reader disagreement about a directory is
  version-independent. A race in `net.Server.listen({fd})` is not.

## Where it does live

`ECONNREFUSED`, not a hang, means the **last descriptor was closed** — the
socket stays in LISTEN while any fd survives, so an unarmed-but-holding standby
would have hung instead. `gap-relay.mjs` has exactly two `process.exit` sites:
one at startup when the parent env is unset, and

```js
srv.on("error", (e) => {
  const held = srv.listening;
  process.stderr.write(...);
  if (!held) process.exit(1);
```

The test asserts the standby is already on the port before the kill, so the
startup one is passed and this is the only survivor: `listen({fd: 3})` failing
at arm time, after node has closed the descriptor.

I could not reproduce it locally in 45 runs, so that is where the evidence
stops. It is a localization, not a proof.

## Why it cannot be proven from a CI log, and why the obvious fix is wrong

That handler writes *"the listen failed, so there is no descriptor left to
keep"* — into a discarded stream. The standby is spawned
`stdio: ["ignore", "ignore", "ignore", fd]`, so its own account of its death
does not survive it. That is why the CI log carries only the wiring banner.

**Giving it an inherited stderr is not the fix**, and the code says why in its
own comment beside that line: a detached child that never closes an inherited
pipe is a pipe the launcher waits on forever — measured at 30 s of a `close`
that never fired. The sibling gap gets `"inherit"` precisely because it dies
with its parent; this one outlives it.

The comment's answer is that what it would have said is on `/health` anyway.
That holds for a standby that is alive to be asked. It does not hold for this
failure, which is the process **exiting** — the one message that explains the
exit is the one that cannot survive it. A destination that does not share a
lifetime with the parent (a file beside the run's state, a line the successor
can pick up) would close that without reintroducing the pipe.

No code change on this branch: adding a retry or widening the assertion would
turn CI green while leaving a real product race — a hard-killed holder can
strand the address the standby exists to protect — buried inside a
test-hygiene PR.

codeslake:
Thanks — the Windows blind spot was real and both halves are fixed at a later head.

Reviewed: `a2834ba0`. Current head: `cf44ba6`.

**The helper sets every key**, not the one this platform happens to read first:

```js
// ALL THREE, because os.tmpdir() reads a DIFFERENT one first per platform:
// TMPDIR || TMP || TEMP on POSIX, TEMP || TMP on Windows.
for (const key of ["TMPDIR", "TMP", "TEMP"]) process.env[key] = FILE_TMP;
```

**The smoke test no longer passes by proving the wrong path.** It was the
sharper half of your finding — an empty `outer` is not evidence when
`os.tmpdir()` never used it — so the probe now carries two guards of its own and
the spawn points all three keys at the inherited directory:

```js
{ env: { ...process.env, TMPDIR: outer, TMP: outer, TEMP: outer } }
```

- **A denominator**: the probe writes a marker from the process AND from a
  grandchild, and `exit(3)` if either is missing. An empty directory now proves
  isolation only when something was actually written somewhere.
- **The key assertion you asked for**: `exit(4)` when any of `TMPDIR`/`TMP`/`TEMP`
  still names the inherited root. On Windows that is what fails rather than
  quietly passing, because `os.tmpdir()` there follows `TEMP` first.

The grandchild is deliberate: the launcher is a CHILD reading its own
`os.tmpdir()`, so a redirection that does not cross the process boundary
isolates nothing.

Measured at head: `node --test test/tmp-isolation.test.mjs` → 1/1 pass.

Could you re-review at `cf44ba6`?

PR #355, opened 2026-08-24
fix(auto-1m-guard): emit the advisory once per process, not per request

codeslake:
## An advisory is advice, and advice does not need repeating

`auto-1m-guard` writes one stderr line when it sees `context-1m-2025-08-07` on an outbound request. The wording is fixed for the mode the proxy runs in, so every copy after the first carried no information — and on a long-lived proxy it crowded everything else off stderr. Measured on one host: **55,685 lines, ~9.8 MiB, about 0.23 per second.**

Latched to the first detection. Per-request detection is unaffected and still observable: the `_auto1mGuard` annotation is written on **every** detected request, spread into the per-session JSON, and read back by the statusline. The latch sits **below** the annotation, and a case pins that ordering.

## The latch has to outlive a module instance

A module-scoped `let` is not once per process. `loadExtensions` cache-busts every import (`+ "?t=" + cacheBuster`), so the module is re-evaluated inside one process on every extension reload, and the latch re-arms with it. Measured through the real `loadExtensions`:

```
module-scoped latch:  one load, 5 requests  -> 1 advisory
                      two loads, 1 request each -> 2 advisories
globalThis latch:     two loads, 1 request each -> 1 advisory
```

The rate that fixes is **once per extension reload** — bounded by how often someone saves a file under `proxy/extensions/`, not by traffic. An earlier revision of this description said "at request rate"; that was wrong and the measurement above is what replaced it.

`Symbol.for`, not `Symbol()`: the registry is what makes every re-evaluated copy reach the same object, and mutating it to `Symbol()` kills the case.

## Also here

`request-capture` declares `routes: ["messages"]` explicitly. `runOnRequest` defaults to exactly that, so the value changes nothing — what it changes is who can see it, since the SCOPE note at the top of that file reasons about the default and an inherited one is invisible to anyone widening the corpus from the export. Both other extensions relying on the same default spell it out.

That file's scope comment was also wrong about which half does the work: the outer half is the **pipeline's** (no `routes` declared → `runOnRequest` skips the hook for every other tagged route), and the body gate scopes an untagged caller. The case now drives the outer half through `runOnRequest` instead of calling `onRequest` directly, where it could not observe the real mechanism.

## Verification

`34/34` across the two touched test files; whole suite green on the rebuilt integration branch.

| reverted | dies |
| --- | --- |
| latch back to a module-scoped `let` | `two module instances in one process wrote 2 advisories` |
| `Symbol.for` → `Symbol` | same case |
| `routes` widened to `["messages","bootstrap"]` | `the bootstrap route reached the capture hook` |
| the latch removed entirely | both advisory cases |

**10 files changed, +140 / −12**, including README, CHANGELOG, the directive, and the two translated READMEs — which still described the pre-latch behaviour and contradicted all three English documents.

## One correction carried in the diff

`classify()` in `proxy-held-port.test.mjs` assumed a body. `node --test` runs files concurrently in CI and alone locally, and `freePort()` releases a port before its caller binds it — so a neighbouring file can hold the number this one just drew, and its 200 arrives where a body was expected. The file was 34/34 locally while `test (22)` was red; that difference was the finding. A non-string is not an `ERR:` body and is not an outage either.


codeslake:
Thanks — the module-instance boundary was the right call, and it is fixed at a
later head.

Reviewed: `c8677fde`. Current head: `5d2159f`.

**The latch is process-durable now**, anchored exactly as you suggested:

```js
// `Symbol.for`, not `Symbol()` -- the registry is what makes every
// cache-busted instance find the same object.
const _latch = (globalThis[Symbol.for("cache-fix.auto-1m-guard")] ??= { advised: false });
```

**Measured with your own boundary probe** — two cache-busted imports in one
process, each driven with a ctx carrying the 1M beta token:

| head | `cacheBustedSameModule` | advisory lines |
| --- | --- | --- |
| `c8677fde` (reviewed) | `false` | **2** |
| `5d2159f` (head) | `false` | **1** |

The modules are still distinct objects — `loadExtensions()` cache-busts as
before — and the advisory is now emitted once for the process regardless.

**The wording is back to the stronger contract** rather than the weaker one the
earlier round had lowered it to. `README.md:618` now reads *"latched to the first
detection for the life of the process (the advice never changes; an extension
reload does not re-arm it)"*, and the directive and CHANGELOG say the same, so
the README no longer under-promises against the code.

`node --test test/proxy-auto-1m-guard.test.mjs` → 25/25 pass at head.

Could you re-review at `5d2159f`?

PR #356, opened 2026-08-24
fix(holder): a handover must not freeze the config it was started with

codeslake:
> **This description covered one of 39 commits.** A reviewer working from the old text reviewed the handover-env re-read and never saw the drain rewrite, which is the larger and riskier half. Rewritten against the current head.

## Two subjects

| | files | size |
| --- | --- | --- |
| the handover env re-read | `bin/handover-env.mjs` (new), `bin/claude-via-proxy.mjs` | +95 / −2 |
| the per-connection drain | `proxy/server.mjs` | +403 / −32 |
| the cases for both | 8 test files | +1950 / −51 |

**13 files, +2444 / −85, 39 commits.** Four fifths of the diff is tests, at roughly 4:1 — there is no refactor in here to peel off. The seam a split would run along is the table above, not code-vs-cleanup.

---

## 1. A handover must not freeze the config it started with

`SIGUSR2` hands the listening socket to a successor spawned from the holder's own environment. That environment is a snapshot of whatever the first launcher was started with, so a handover carried stale gate values forward indefinitely — the proxy kept serving with settings the operator had already changed.

`bin/handover-env.mjs` reads them back from a file at handover time. The holder still **pins** five keys over the file, because they describe *this* process and not the operator's intent: a successor that inherited them would claim a port it does not hold.

A key **absent** from the file is inherited, not unset. `CACHE_FIX_STANDBY: undefined` really drops the key — measured against the control (`{"present":true,"value":"1"}` inherited vs `{"present":false}` pinned, with the parent holding it set).

The file is honoured only once its trailing newline is written, so a half-written line is never read as a setting.

## 2. A drain ends when the bytes stop, not when a clock does

A stop used to cut every in-flight reply at a fixed 5 s ceiling. Measured on one host: **15 replies severed across four stops** (4, 3, 1, 7), every one mid-response, with no stall predicate installed at all.

The drain is now per connection. A connection that has written a byte within `CACHE_FIX_DRAIN_STALL_MS` (default 90000) keeps the drain alive; one that has gone silent that long is ended alone. `CACHE_FIX_DRAIN_MS` (default 1800000) is a **backstop and a re-evaluation point**, not a deadline — a connection still moving bytes when it expires survives it and is reported every 60 s.

The 5 s ceiling is kept for exactly one arm: a **standalone** stop, where the process draining IS the process a service manager is waiting on. The predicate is one line — `handedOff || handoverRelease || heldByLiveHolder` — and each term is measured in the code's own comment.

Other changes in this half: the listening socket is unbound at the **net layer** immediately, so the port is free while the drain runs; the forced-close line reports **silence**, not the age of the request; a reply that ended itself is counted as owed but not as a cut; and `SIGUSR2` replaces `SIGHUP` for the handover so a plain stop and a handover stop being told apart no longer depends on a log line.

## 3. Three defects this branch fixed in itself

- **A negative stall budget cut everything.** `Number(x) || fallback` took the fallback for an explicit `0` and passed a **negative** through, and `now - at < -1` is false on the first tick — the guillotine this drain replaced, one typo away.
- **Then whitespace did the same.** `Number(" ")` is `0`. Measured end to end: `CACHE_FIX_DRAIN_STALL_MS=" "` severs a healthy 3-second-gap stream that survives at any real value. Both knobs now take a plain non-negative decimal.
- **The route tally rendered an authority.** `drainRoute` guarded on the request-target's SHAPE, and a shape test cannot enforce a content rule: `/http://user:pass@host/v1`, a backslash, a percent-encoded `//`, a `;`-prefixed authority and a `#` fragment are each one leading slash. It judges the label it is about to write now — no `@`, no `#`.

## Verification

Whole suite **1954 → 2006 tests, 2005 pass, 0 fail, 1 skipped**, and the rebuilt integration branch carrying this plus four sibling PRs is green on the same count.

| reverted | dies |
| --- | --- |
| the decimal test in `drainBudgetMs` | `whitespace " " was read as an explicit budget` |
| the `[@#]` filter in `drainRoute` | `an authority-first target rendered into the log: /http:/user:...` |
| the release comparison forced to disagree | eight cases, by name |

The lifted-source case that evaluates the budget line is handed the **real** helper rather than a copy, for the reason its own comment already gives about `unwaited`: a re-implementation lets the two drift into agreeing about a budget the server does not use.

## Known, and not fixed here

**A resident drainer holds the launcher's inherited stderr for the whole drain.** The holder settles on the release line and exits in milliseconds; the drainer keeps fd 2, so anything **capturing** `run-service`'s output — a CI step, `$(… 2>&1)`, `ssh host '…'` — waits out the drain. Measured, with a control:

```
held arm (long budget):   no stderr EOF and no exit 20s after SIGTERM, exitCode still null
standalone arm (5s):      stderr EOF at +5011ms
```

A terminal is unaffected; only a pipe blocks. Every available fix trades the drain's live reporting — which four of this PR's own cases read — against the hang, so it wants its own change rather than a patch tacked on here.


codeslake:
Thanks for the crash detail — the exit code and the stderr size are more useful than a timeout would have been.

I measured the diff before answering, because the split as proposed rests on there being ~2000 lines of refactor, and there is not:

```
total          +2444 / -85    13 files, 39 commits
proxy/          +388 / -32     1 file
bin/             +95 /  -2     2 files
test/          +1950 / -51     8 files      <- 80% of the diff
docs/README/CHANGELOG   0 / 0               <- a real gap, now fixed
```

So there are no renames, moves, or structural cleanup to peel off. The other ~2000 lines are tests for the contract change, at roughly 4:1 — which `AGENTS.md` explicitly says not to flag. Splitting them from the production change they exercise would make both halves less reviewable, not more.

A split **is** available, just along a different seam. The body describes one of the 39 commits (the handover-env re-read); the rest is a per-connection drain rewrite — the net-layer unbind, `SIGUSR2` replacing `SIGHUP`, the holder settling on the release line, the stall predicate, `drainRoute`, and the two budget knobs. Those are two subjects, and separating them is coherent. It also means rewriting 39 commits into two branches and discarding the review rounds already done against this head, so I would rather not do it on my own initiative.

The cheaper thing, if it helps your side: **the production diff is 483 lines across 3 files.** If a one-shot review can take that with the tests available as reference rather than as input, that covers every semantic change in the PR. `git diff 78ca9483..HEAD -- proxy/ bin/` is the whole of it.

Two things I owe regardless, and am doing:

- The body describes 1 of 39 commits. A reviewer working from it reviews the wrong diff — which is exactly the failure your R1 focus items were pointing at. Rewriting it against the current head.
- Three defects landed in the last two rounds, all in code I wrote and all found by review rather than by the tests: a stall budget that read whitespace as an explicit zero (measured severing a healthy 3-second-gap stream), a route label that still rendered `user:pass@host` through five shapes a leading-slash test admits, and before those, a reaper that could unlink a fingerprint record published inside its own probe. Fixed and pushed.

Happy to take the split if you'd rather have it — say the word and I'll cut it along the handover-env / drain line.

